PR #25, opened 2023-11-19
Support BcIr::Optcase.

sisshiki1969:
Support optimized case statements in both of VM and JIT.
This converts an if-then-chain to a single table jump in the case that condition expression for all of  `when` branches were integers.


sisshiki1969:
## benchmark

before:
#### Iteration per second (i/s)
|      |3.3.0-preview3| 3.3.0-preview3 --yjit|../target/release/monoruby|../target/release/monoruby --no-jit|
|:-----|-------------:|---------------------:|-------------------------:|----------------------------------:|
|case  |       106.548|                98.197|                   157.341|                             45.912|

after:
|      |3.3.0-preview3| 3.3.0-preview3 --yjit|../target/release/monoruby|../target/release/monoruby --no-jit|
|:-----|-------------:|---------------------:|-------------------------:|----------------------------------:|
|case  |       106.810|                98.567|                   378.443|                            264.381|

Issue #177, opened 2026-03-22
instance_eval block does not resolve constants in receiver's scope


sisshiki1969:
## Summary

When `instance_eval` is called with a block, constants defined in the receiver's class are not accessible. Instead, a `NameError: uninitialized constant` is raised.

## Failing ruby/spec test (1 error)

From `core/basicobject/instance_eval_spec.rb`:

- **doesn't get constants in the receiver if a block given** — `NameError: uninitialized constant X_BY_BLOCK`

## Expected behavior

Constants defined in the receiver's class should not be accessible from `instance_eval` block context (the test expects `NameError` to be raised, but for a different reason than what happens).

The actual error path differs from CRuby's behavior, indicating a constant resolution issue in the `instance_eval` block scope.

## How to reproduce

```
cd ../spec && ../mspec/bin/mspec core/basicobject/instance_eval -t monoruby
```

PR #181, opened 2026-03-22
Fix rescue modifier to wrap only RHS of assignments

sisshiki1969:
## Summary
- Fix rescue modifier parsing so it wraps only the RHS of assignment expressions, not the entire assignment
- Refactor to handle rescue modifier during RHS parsing instead of post-hoc wrapping
- Add `parse_arg_with_rescue` and `wrap_mrhs_with_rescue` helpers for consistent rescue handling across assignment forms

## Test plan
- [x] `cargo test` passes
- [x] Verify `x = expr rescue value` correctly wraps only `expr` in begin..rescue
- [x] Verify compound assignments (`+=`, `||=`, etc.) also handle rescue correctly

### DIFF
--- a/monoruby/tests/rescue.rs
+++ b/monoruby/tests/rescue.rs
@@ -315,6 +315,108 @@ fn retry_in_loop() {
 }
 
 #[test]
+fn rescue_modifier_simple_assign() {
+    run_test(
+        r#"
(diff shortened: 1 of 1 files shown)


sisshiki1969:
close #177 .

PR #373, opened 2026-04-29
jit: virtual BOOL_CLASS for inline-cache stability on bool flips

sisshiki1969:
## Summary

Introduces a virtual `BOOL_CLASS` (id 54) that exists only inside inline
caches and JIT type predictions. `true.class` is still `TrueClass`,
`false.class` is still `FalseClass`, and `TrueClass.superclass` is still
`Object` — the new class id has no Ruby-level identity.

## Why

Profiling `optcarrot --benchmark` (180 frames) under `--features profile`
showed three hot sites whose receivers toggle between `true` and `false`,
each tripping the per-receiver-class guard on every flip and re-entering
the deopt-recompile loop:

```
( 1121 )  PPU#poke_2006   [:00006]  %3 = !%3   [TrueClass]
(   86 )  PPU#main_loop   [:00153]  %3 = !%3   [FalseClass]
(   81 )  PPU#setup_frame [:00005]  %1 = !%1   [TrueClass]
```

`@scroll_toggle = !@scroll_toggle` and similar idioms generate a steady
~1300 deopts across 180 frames — small per-flip but a constant
recompilation source.

## How

The contract is: **lookups under `BOOL_CLASS` resolve to the same
`FuncId` regardless of which boolean the receiver was**. Two pieces
together enforce it:

1. **Same-FuncId mirroring for shared methods.** The boolean operators
   `^`, `|`, `&` were previously written twice in pure Ruby (one set
   per class) so they had distinct `FuncId`s. They move to a tiny Rust
   module `bool_class.rs` that registers each operator **once** and
   calls `add_method` on both `TrueClass` and `FalseClass` with the
   same `FuncId`. The bodies branch on `lfp.self_val().as_bool()` to
   recover the per-receiver semantics.

2. **Lookup-time invariant check.** `Store::check_method_for_class`
   recognizes `BOOL_CLASS` and walks both `TrueClass` and `FalseClass`,
   only returning an entry when their `FuncId`s match. If user code
   defines a method on only one of the two classes the lookup fails,
   and `Executor::find_method` falls back to the per-class lookup so
   the IC stays per-class.

Plumbing wired through:

- `Value::class_for_ic()` — returns `BOOL_CLASS` for booleans, otherwise
  `class()`.
- Runtime asm `get_class` — `or rdi, 8 ; cmp rdi, TRUE_VALUE`. Both
  `TRUE_VALUE` (0x1c) and `FALSE_VALUE` (0x14) collapse to 0x1c after
  the OR; no other tagged value lands at 0x1c. The bytecode IC therefore
  records `BOOL_CLASS` for any boolean receiver.
- JIT class guard for `BOOL_CLASS` uses the same OR-then-cmp test (4
  instructions, scratches `rax`).
- `Guarded::from_concrete_value` uses `class_for_ic`, so a literal
  boolean in a `LinkMode::C` slot reads back as
  `Guarded::Class(BOOL_CLASS)`. `is_truthy` / `is_falsy` short-circuit
  on the literal value; non-literal slots guarded as `BOOL_CLASS`
  conservatively return false (truthiness is unknown at compile time).

## Spec impact

`cargo test --release --lib`:
- **before**: 1404 passed
- **after**: 1439 passed (+35 from now-shared `^/|/&` coverage)

The same two pre-existing failures (`string_inspect`,
`symbol_inspect_unquoted` — non-ASCII inspect formatting under Ruby 4)
remain. `codegen::tests::guard_class` was updated to expect
`BOOL_CLASS` for boolean inputs.

The targeted deopts disappear from the top-20 list:

```
before: PPU#poke_2006 1121 / PPU#main_loop 86 / PPU#setup_frame 81
after:  none of these appear
```

## Optcarrot benchmark (180 frames, 8 trials)

|  | avg fps | range |
|---|---:|---|
| master baseline | 188.6 | 175 – 192 |
| this PR | 187.9 | 181 – 196 |

FPS difference is within noise; the change's primary effect is the
elimination of ~1300 deopts on `!@flag`-style sites, which removes a
recompilation source and stabilizes JIT code across frames.

## Test plan

- [x] `cargo build` clean
- [x] `cargo test --release --lib` — 1439 passed, same 2 pre-existing
      `string_inspect` / `symbol_inspect_unquoted` failures, no new ones
- [x] Optcarrot --benchmark — checksum 59662 matches CRuby's
- [x] `monoruby -e 'p TrueClass.superclass; p true.class; p true.is_a?(TrueClass)'`
      → `Object`, `TrueClass`, `true` (Ruby semantics preserved)

https://claude.ai/code/session_01SRrCAdBwQ5RbixCCtNwYtD

---
_Generated by [Claude Code](https://claude.ai/code/session_01SRrCAdBwQ5RbixCCtNwYtD)_

### DIFF
--- a/monoruby/builtins/builtins.rb
+++ b/monoruby/builtins/builtins.rb
@@ -42,33 +42,10 @@ def caller_locations(start = 1, length = nil)
   end
 end
 
-class TrueClass
-  def ^(other)
-    !other
-  end
-
-  def |(other)
-    true
-  end
-
-  def &(other)
-    !!other
-  end
-end
-
-class FalseClass
-  def ^(other)
-    !!other
-  end
-
-  def |(other)
-    !!other
-  end
-
-  def &(other)
-    false
-  end
-end
+# `^`, `|`, `&` for `true` and `false` live on the internal `Boolean`
+# parent class so that `true.method(:&) == false.method(:&)` and the
+# JIT inline cache can treat the receiver as `BOOL_CLASS` regardless of
+# which boolean was observed first.
 
 class NilClass
   def ^(other)
--- a/monoruby/src/builtins.rs
+++ b/monoruby/src/builtins.rs
@@ -2,6 +2,7 @@ use super::*;
 
 mod array;
 mod binding;
+mod bool_class;
 mod class;
 mod dir;
 mod encoding;
@@ -59,6 +60,7 @@ pub(crate) fn init_builtins(globals: &mut Globals) {
     true_class::init(globals);
     false_class::init(globals);
     nil_class::init(globals);
+    bool_class::init(globals);
     module::init(globals);
     class::init(globals);
 
--- a/monoruby/src/builtins/bool_class.rs
+++ b/monoruby/src/builtins/bool_class.rs
@@ -1,0 +1,88 @@
+use super::*;
+
+//
+// Boolean methods (internal): `BOOL_CLASS` is *not* a Ruby-visible class.
+// `true` and `false` keep their normal `TrueClass` / `FalseClass`
+// identities at the user level. The constant only exists as an inline-
+// cache tag so a polymorphic-on-true/false call site does not deopt on
+// every flip.
+//
+// To make that tag safe, the methods that are otherwise defined twice
+// (once on TrueClass and once on FalseClass with subtly different bodies)
+// are registered **once** in Rust and `add_method` is called on **both**
+// classes with the same `FuncId`. Subsequent JIT lookups under either
+// class therefore resolve to the same `FuncId`, which is the invariant
+// the IC relies on.
+//
+
+pub(super) fn init(globals: &mut Globals) {
+    define_bool_method(globals, "&", and_, 1, 1, false);
+    define_bool_method(globals, "|", or_, 1, 1, false);
+    define_bool_method(globals, "^", xor_, 1, 1, false);
+}
+
+/// Register `address` once and add it as `name` to both `TRUE_CLASS` and
+/// `FALSE_CLASS` with the **same** `FuncId`.
+fn define_bool_method(
+    globals: &mut Globals,
+    name: &str,
+    address: BuiltinFn,
+    min: usize,
+    max: usize,
+    rest: bool,
+) {
+    let func_id = globals
+        .store
+        .new_builtin_func(name, address, min, max, rest, &[], false);
+    globals.gen_wrapper(func_id);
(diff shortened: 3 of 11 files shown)


sisshiki1969:
## Follow-up: fix `bin/emit` core dump on `so_nbody.rb`

Pushed [`981cb4a5`](../commit/981cb4a5) on top of this PR.

### Symptom
`bin/emit` (which builds `--features emit-asm` and dumps disassembly for each benchmark) was core-dumping on `benchmark/so_nbody.rb`:

```
thread 'main' panicked at monoruby/src/globals/store/class.rs:451:21:
called `Option::unwrap()` on a `None` value
...
   5: ClassInfo::get_module
   6: Store::debug_class_name
   7: TraceIr::format
   8: Codegen::dump_disas
  ...
  19: jit_compile_patch                  <-- extern "C"
thread caused non-unwinding panic. aborting.
```

### Cause
The virtual `BOOL_CLASS` (`ClassId(54)`) introduced by this PR has `ClassInfo.object: None` by design — it's IC-only and has no Ruby-visible `Module`. But `Store::debug_class_name` eagerly called `ClassInfo::get_module()` (which `unwrap()`s `self.object`) before checking whether the class had a name.

When `dump_disas` formatted a `TraceIr::UnOp` whose IC had cached `BOOL_CLASS` (e.g. an `!@flag`-style site), the `unwrap` panicked. Because the panic originated inside `extern "C" jit_compile_patch`, it could not unwind and the process aborted.

The site that actually triggered it in `so_nbody.rb` shows up in the disas as:
```
%2 = !%2                             [BOOL]
```

### Fix
- Add `ClassInfo::try_get_module() -> Option<Module>`.
- Route `Store::debug_class_name` through `try_get_module()`. Classes with no Module fall back to `format!("{:?}", class)`, which the existing `Debug` impl already renders as `BOOL` for `ClassId(54)`.

### Verification
After the fix, all 11 benchmarks in `bin/emit` complete without aborting:

| benchmark | before | after |
|---|---|---|
| `app_fib`, `tarai`, `app_aobench`, `so_mandelbrot`, `binarytrees`, `plb2/{nqueen,sudoku,matmul,bedcov}` | exit 0 | exit 0 |
| **`so_nbody`** | **SIGABRT (core dumped)** | **exit 0** |
| `quick_sort` | exit 1 (pre-existing stack overflow, unrelated) | exit 1 |

Disassembly now correctly renders `[BOOL]` as the IC tag for unary-op sites whose receiver oscillates between `true`/`false`.

PR #409, opened 2026-05-05
Prism parser backend (default) + assorted perf and GC fixes

sisshiki1969:
## Summary

Bring up a new **Prism-based parser backend** for monoruby and make it the default, with the legacy `ruruby-parse` backend kept as an opt-in fallback (`MONORUBY_PARSER=ruruby`). The two backends produce the same `crate::ast::ParseResult`, so bytecodegen and below are unchanged. Also bundles a batch of unrelated perf, GC-correctness, and YAML-parser fixes that landed alongside the parser work.

### Parser changes

- New module [parser/](monoruby/src/parser/) with `Backend::{Prism, Ruruby}` selectable at runtime via `MONORUBY_PARSER`. Default is `Prism`.
- `prism_backend.rs` (~3,650 lines) lowers ruby-prism's AST into monoruby's existing AST. Phased rollout across 25+ commits with full coverage of: locals/control-flow/collection literals, blocks/methods/classes/modules/constants, ivars/gvars/parameters, string interpolation, begin/rescue/else/ensure, multi-assign and op-assign, lambda/&proc/index/attr op-assign, const-path/super/yield/case/for/jump/defined/alias, regex/singleton-def/numerics/pseudo-vars, etc.
- ruby-prism is **vendored** under [vendor/ruby-prism/](vendor/ruby-prism/) so we can extend it with `parse_with_options` for `eval` / `binding.eval` (passing surrounding scope locals).
- Unsupported nodes raise a `MonorubyErr::FatalError` pinned to the offending source line, instead of silently falling back or panicking.

### Other changes on this branch

- **GC**: fix mark-leak bugs in `Array#zip / #grep / #uniq / #&` etc. (and `Set#collect!`) by rooting fresh values on `temp_stack` instead of relying on disabled GC. Reproducible with `--features gc-stress`.
- **Array perf**: O(n+m) `Array#&`, `#-`, `#intersection` via RubySet; 1-pass `Array#minmax` with all-Fixnum tight i64 fast path; Fixnum fast path on `Array#sum` and `Array#inject(:+)`.
- **String perf**: cache merged `CodeRange` across `<<` / `extend_from_slice_checked` / `set_byte` / `bytesplice` to drop O(n²) into O(n); UTF-8 prefix-free fast path for `bytesplice` skips re-validation; `Encoding#ascii_only?` uses cached cr instead of `[u8]::is_ascii()`.
- **Hash**: port `Hash#initialize` from Ruby to Rust.
- **Psych** (pure-Ruby YAML parser): byte-walking versions of `skip_blanks_and_comments` / `indent_of` / `block_mapping_line?`, multi-line quoted strings, full block-scalar indicators (`|N`, `|N-`, `|2+`), correct chomp semantics, quoted keys, inline-comment handling. Output now byte-identical to CRuby on `discourse_client.en.yaml`.
- **Test infra**: `bin/ruby-bench-diff` (exits non-zero on output mismatch against CRuby for erubi / etanni / fannkuchredux / psych-load / blurhash); wired into `bin/test` and into [.github/workflows/rust.yml](.github/workflows/rust.yml) (clones ruby-bench, installs erubi gem). Ruby version pin bumped to 4.0.1 in CI; locally `find_ruby` skips too-old `ruby` on PATH and prefers rbenv ≥ 4.0.

### Compatibility

- Default behavior changes: monoruby now parses with Prism by default. Set `MONORUBY_PARSER=ruruby` to fall back.
- The two parsers are interchangeable for bytecodegen — output `ParseResult` is identical in shape.
- All existing integration tests, optcarrot output diff, and `bin/ruby-bench-diff` output diff pass on Prism (CI run on this branch is green).

## Test plan

- [x] `bin/test` (cargo nextest with `gc-stress` + `stress-spill-pool`, benchmark output diffs, optcarrot diff, `bin/ruby-bench-diff`)
- [x] Output-diff against CRuby/YJIT for: erubi, etanni, fannkuchredux, psych-load, blurhash
- [x] `bin/spec` — ruby/spec on monoruby (no regressions vs master)
- [x] Reviewer: spot-check `MONORUBY_PARSER=ruruby` still works for any script that worked before


sisshiki1969:
## Parser performance: prism vs ruruby-parse

All numbers are **best-of-5** wall-clock seconds from `/usr/bin/time -f %e`, on the release binary (`cargo build --release -p monoruby`), x86-64 Linux (WSL2).

### Parse + bytecodegen only

Source file is concatenated 30× into a single blob and `exit 0` is prepended, so monoruby parses + bytecodegens the entire blob and exits before executing. Removes both runtime cost and the AST-dump format-cost distortion of `--ast`.

| File (effective lines)                    | prism | ruruby | speedup |
|-------------------------------------------|-------|--------|---------|
| optcarrot/lib/optcarrot/ppu.rb (~44,040L) | 0.55s | 0.67s  | **1.22x** |
| optcarrot/lib/optcarrot/cpu.rb (~34,860L) | 0.33s | 0.44s  | **1.33x** |
| optcarrot/lib/optcarrot/apu.rb (~25,680L) | 0.32s | 0.41s  | **1.28x** |
| monoruby/stdlib/psych.rb (~18,330L)       | 0.12s | 0.19s  | **1.58x** |
| benchmark/app_aobench.rb (~8,790L)        | 0.08s | 0.13s  | **1.62x** |

Both backends pay an identical ~60 ms of monoruby fixed startup (load_path setup, etc.) and an identical bytecodegen cost; the delta is the parser. Prism is **1.2x–1.6x faster** at parsing real Ruby across this set.

### End-to-end (parse + bytecodegen + JIT + execute)

| Script                       | prism | ruruby | speedup |
|------------------------------|-------|--------|---------|
| benchmark/app_fib.rb         | 0.11s | 0.14s  | 1.27x   |
| benchmark/tarai.rb           | 0.12s | 0.14s  | 1.17x   |
| benchmark/so_nbody.rb        | 0.16s | 0.19s  | 1.19x   |
| benchmark/binarytrees.rb     | 0.29s | 0.35s  | 1.21x   |
| optcarrot -b --opt --debug   | 7.94s | 8.32s  | 1.05x   |

For short-running scripts the parser saves ~30–60 ms outright (proportionally large because the script itself runs in well under 100 ms). On long-running optcarrot the saving is the same absolute parse cost (~0.4 s out of ~8 s).

### Notes

- A naive `--ast prism` vs `--ast ruruby` measurement looks like a 22–43× win for prism, but ~33× of that is just the format/IO cost of dumping a 6.8 MB `Annot{...}` debug string vs. prism's 0.2 MB compact dump. The numbers above use `exit 0` instead, which pays no AST-dump cost.
- ruruby-parse remains the fallback for `eval` / `binding.eval` only when explicitly opted into via `MONORUBY_PARSER=ruruby`. Default end-to-end runtime has no ruruby-parse on the hot path anymore.
- Methodology / scripts: `/tmp/bench_parsers3.sh` (parse-only) and `/tmp/bench_optcarrot.sh` (end-to-end) on this branch's HEAD ([e5365adb](https://github.com/sisshiki1969/monoruby/commit/e5365adb)).

PR #412, opened 2026-05-06
Switch parser to Prism

sisshiki1969:
## Summary

Brings the Prism parser backend onto master in 37 commits: the `monoruby::ast` boundary, a vendored `ruby-prism` 1.9.0 with `parse_with_options(source, &Options)` for eval/binding scope-locals, the lowerer, and the targeted fixes that brought Prism to feature-parity with the existing ruruby-parse path.

### Phase 0–1a (parser-AST boundary)
- `prism phase 0`: add `ruby-prism` dependency and `--dump-ast=prism`
- `prism phase 1a`: introduce `monoruby::ast` as the parser-AST boundary (decouple bytecodegen / executor from the parser AST type)

### Phase 2 (lowerer coverage, in batches)
- locals / control flow / collection literals
- blocks, methods, classes, modules, constants
- return, ivars, gvars, full parameter coverage
- string interpolation + `begin/rescue/else/ensure`
- multi-assign and op-assign coverage
- destructure / lambda / `&proc` / index / attr op-assign
- const-path / super / yield / case / for / jump / `defined?` / alias
- regex / singleton-def / numerics / pseudo-vars / op-asgn rest

### Switching backends + targeted fixes
- vendor `ruby-prism` 1.9.0 with `parse_with_options` so eval/binding can carry scope-locals
- pin eval/binding to ruruby until the vendored API was wired up; then move them onto Prism
- switch default backend to Prism; `MONORUBY_PARSER=ruruby` opts back out per-test
- assorted node-coverage fixes (regex flag handling, postfix `while/until`, splat-on-RHS multi-assign, `is_safe_navigation`, bare-splat `return/break/next`, mixed-key trailing hash arg through `hash_splat`, `MatchWriteNode`, `CallTargetNode` in multi-write LHS, forwarding super with literal block, toplevel `class ::Foo`, eval `line_offset`)
- replace silent ruruby-fallback with hard panic on lowerer gaps, then surface those gaps as `FatalError` at the offending line instead of `panic!`
- drop `LowerError`, raise `MonorubyErr` directly from the lowerer

### Final API shape (HEAD)
- `parse_with_options<'pr>(source: &'pr [u8], options: &Options) -> ParseResult<'pr>` — `pm_parser_init` copies everything it needs out of the options block (scope locals are `xmalloc`/`memcpy`'d into the parser's own scope stack), so the borrow ends with the call and no keep-alive field is needed on `ParseResult`. `Options` is a `#[repr(transparent)]` newtype around `pm_options_t` whose libc-heap scope/local storage is freed in `Drop` via `pm_options_free`.

## Test plan

- [x] `cargo build -p monoruby`
- [x] `cargo test -p monoruby --test variables --test method_call --test rescue` (63 + 35 + 13 pass under `MONORUBY_PARSER=prism`)
- [ ] CI: full `bin/test` run on this branch


tekknolagi:
Whoa, why the switch?

sisshiki1969:
I’m a bit tired of maintaining my own Ruby parser, so I’m going to rely on Prism from now on! :)

https://x.com/s_isshiki1969/status/2051626941304983582?s=46&t=gZ9WqyHwywijMDrpk0xfTQ

tekknolagi:
Neat, thanks

PR #690, opened 2026-06-07
fix(integer): promote Integer#<< value overflow to Bignum (#687)

### LINKED ISSUE #687: Integer#<< by an amount < 64 wraps instead of promoting to Bignum (safe_int_shl overflow check is wrong)
## Summary

`Integer#<<` (left shift) produces a **wrapped i64** instead of promoting to a `Bignum` when the result overflows but the shift amount is `< 64`. This is an interpreter bug in `safe_int_shl` (`executor/op/binary_ops.rs`); it affects both the literal and the variable-operand forms, and is independent of the JIT (reproduces with `--features no-jit`).

## Reproduction

```ruby
p 5 << 62        #=> monoruby: 4611686018427387904   CRuby: 23058430092136939520
a = 5; b = 62
p a << b         #=> monoruby: 4611686018427387904   CRuby: 23058430092136939520
p(-256 << 62)    #=> monoruby: 0                      CRuby: -1180591620717411303424

p 1 << 100       #=> 1267650600228229401496703205376 (correct — shift amount >= 64)
```

- `5 << 62` = 5·2^62 = `23058430092136939520`, which exceeds `i64::MAX`, so it must be a Bignum. monoruby returns `4611686018427387904` (= `(5 << 62) mod 2^64` = 2^62).
- `-256 << 62` overflows to a large negative Bignum; monoruby returns `0`.
- `1 << 100` works because the shift amount (100) is `>= 64`, taking a different branch.

`(5 << 62).class` is reported as `Integer` (the wrapped value), not the correct Bignum value.

## Root cause

`safe_int_shl` in `monoruby/src/executor/op/binary_ops.rs`:

```rust
let rhs = rhs as u32;
Some(match lhs.checked_shl(rhs) {
    Some(res) if lhs.is_positive() == res.is_positive() => Value::integer(res),
    _ => bigint_shl(&BigInt::from(lhs), rhs),
})
```

Two problems:

1. **`i64::checked_shl` does not detect value overflow.** It only returns `None` when the shift *amount* is out of range (`rhs >= 64`); for `rhs < 64` it returns `Some(lhs << rhs)` with the high bits silently dropped. So `5i64.checked_shl(62)` is `Some(4611686018427387904)`, not `None`.

2. **The `lhs.is_positive() == res.is_positive()` guard is insufficient.** It only catches overflows that flip the sign. Overflows that keep the sign slip through:
   - `5 << 62`: lhs `+`, res `+` (2^62 < 2^63) → guard passes → wrong.
   - `-256 << 62`: lhs `-`, res `0` (both not-positive) → guard passes → wrong.

## Suggested fix

Detect lost bits with a shift-back comparison instead of relying on `checked_shl` + sign:

```rust
let rhs = rhs as u32;
let res = lhs.wrapping_shl(rhs);
Some(if res.wrapping_shr(rhs) == lhs {
    Value::integer(res)        // no significant bits lost
} else {
    bigint_shl(&BigInt::from(lhs), rhs)
})
```

(This is the same overflow test the aarch64 JIT now uses in `gen_shl`/`gen_shl_rhs_imm`; note the JIT path is correct because it *deopts* to this interpreter on overflow, so the JIT result currently matches the buggy interpreter — `JIT == VM`.)

## Notes

- Found while porting the variable-amount shift inliners to the aarch64 JIT (PR #686). Out of scope there; filing separately.
- Worth adding `run_test` coverage for `n << k` across the i64-overflow boundary (e.g. `5 << 62`, `-256 << 62`, `(1 << 40) << 30`) once fixed, since these currently diverge from CRuby.

sisshiki1969:
Fixes #687.

## Problem

`Integer#<<` by an amount `< 64` wrapped into a bogus i64 fixnum instead of promoting to Bignum when the value overflowed:

| expression | before | CRuby |
|---|---|---|
| `5 << 62` | `4611686018427387904` | `23058430092136939520` |
| `-256 << 62` | `0` | `-1180591620717411303424` |
| `3 << 62` (const-fold) | `-4611686018427387904` | `13835058055282163712` |

## Root cause

Two helpers used `i64::checked_shl`, which only reports failure when the **shift amount** is `>= 64` — it silently drops the high bits on a **value** overflow:

- `safe_int_shl` (`executor/op/binary_ops.rs`) — the runtime path. Its `lhs.is_positive() == res.is_positive()` guard only catches sign-flipping overflows, missing cases like `5 << 62` (stays positive) and `-256 << 62` (wraps to 0).
- `fold_shl_pos` (`builtins/numeric/integer.rs`) — the JIT constant-fold path. Its `is_i63(result)` guard accidentally rejected most wrapped values, but a brute-force sweep found 207 inputs where the wrapped value still passed `is_i63`, yielding a wrong fold (e.g. `3 << 62`, `4 << 62`).

## Fix

Detect lost bits with a shift-back test — `(lhs << k) >> k == lhs` — in both helpers. Anything inexact (or, in the folder, not fitting an i63 fixnum) falls through to the Bignum path.

## Testing

- New runtime-path regression test `integer_shl_overflow_promotes_to_bignum`.
- Extended `integer_shift_constant_folding` with the overflow cases.
- Verified against CRuby across an exhaustive sweep of operands × shift amounts (0–70, plus large) on **both** the runtime and constant-fold paths — all match.
- Full `numeric` suite (128 tests) passes.

### DIFF
--- a/monoruby/src/builtins/numeric/integer.rs
+++ b/monoruby/src/builtins/numeric/integer.rs
@@ -853,8 +853,13 @@ fn fold_shl_pos(lhs: i64, k: u64) -> Option<i64> {
         // when k > u32::MAX (handled by safe_int_shl as RangeError).
         return None;
     }
-    let result = lhs.checked_shl(k as u32)?;
-    if Value::is_i63(result) {
+    // `checked_shl` only guards the shift amount (k >= 64), not the value
+    // overflow `lhs << k` silently drops the high bits (e.g. `3 << 62`).
+    // Confirm the shift is lossless by shifting back before trusting it, then
+    // require the result to fit an i63 fixnum; otherwise fold is not possible.
+    let k = k as u32;
+    let result = lhs.wrapping_shl(k);
+    if result.wrapping_shr(k) == lhs && Value::is_i63(result) {
         Some(result)
     } else {
         None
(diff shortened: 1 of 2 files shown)


sisshiki1969:
Fixes #687.

PR #744, opened 2026-06-20
jit(§5): lower AsmInst::Inline to LIR + loop-aware FP allocator (phys-loop-aware)

sisshiki1969:
## Summary

Two steps of the §5 *regalloc separation* effort, on top of the layer2 work already on `master` (#743). The net diff vs `master` is small (7 files, +253/-28); the bulk of the branch history is the already-merged layer2/§39 commits.

### 1. `LInst::Inline` — every `AsmInst` now lowers to LIR
Adds `LInst::Inline(InlineProcedure)`, closing the last `AsmInst` arm that bypassed LIR in the `compile_asmir` dispatcher. After this, **every** `AsmInst` reaching the dispatcher lowers to an `LInst`.

`LInst::Inline` is the one LIR op whose emit is *not* store-free: its wrapped generator closure needs the compile context (`&Store`, `&SideExitLabels`, frame `base`), which the store-free `encode_linst` does not receive. It is therefore dispatched at the AsmIR→LIR lowering boundary via a new `encode_linst_inline`, and hits the `unreachable!` fallthrough if it ever reaches `encode_linst`/`encode_linst_macro`.

`LInst`/`Lir` drop their (unused) `Clone` derive, since `InlineProcedure` wraps a move-only `Box<dyn FnOnce>`.

This is byte-identical / behaviour-preserving — it only re-routes the existing inline-builtin generators through a carrier op so the "all `AsmInst` → `LInst`" invariant holds.

### 2. `phys-loop-aware` — loop-aware FP allocation (default-off, §42)
A new default-off feature implementing §27.3 Stage 2a+2b. The back-edge fixpoint (①) records the loop-carried float set `L`; the phase-1 spill-victim policy then keeps `L`'s `Sf` caches resident (a fresh value spills instead of evicting a loop-carried float), cutting in-loop reloads.

This cracks §29's "inertness": supplying `L` at merge time (where the fixpoint has already computed the back-edge) makes the policy actually fire. Verified non-inert under `stress-spill-pool` (POOL=2: placement count 28→26, digest changes) and inert under the shipping POOL=14 (so shipping is byte-identical). It is **non-byte-identical by design** (§26) — the `shadow-placement` digest becomes a delta meter, not an equality check — so it stays behind the flag until the M1 perf A/B gate clears.

### Docs
- `doc/regalloc_separation.md`: §42 (Stage 2a+2b landing) and §43 (the deferred global-pin lever — headroom measured, the win is copy coalescing, partial).
- `doc/lir.md`: §6/§8 updated for the `LInst::Inline` carrier.

## Verification
- x86_64: `cargo build` clean; **1716 lib unit tests pass** (exercises sqrt / array-index inline builtins through the new `LInst::Inline` path).
- aarch64: `cargo check --target aarch64-unknown-linux-gnu` compiles.
- Correctness under `phys-loop-aware` + `stress-spill-pool`: 2090/2090.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CF87ULUvuPqofXDRcBTcYy)_


sisshiki1969:
Superseded by #745, which is the same change rebased cleanly onto current `master` (this branch had a merge conflict from #743's squash-merge). Closing in favor of #745.

---

PR #825, opened 2026-07-05
bytecodegen: evaluate op-assign receiver / index only once

sisshiki1969:
## Summary

`recv.attr op= v` and `recv[i] op= v` — for arithmetic (`+=`, `-=`, …) and the short-circuit `||=` / `&&=` — evaluated the receiver (and, for an index target, each index argument) **twice**: once for the getter read and once for the setter write. CRuby evaluates them exactly once.

- Arithmetic op-assign lowers to `AssignOp(op, target, value)`, and bytecodegen read `target` twice (`eval_lvalue` for the setter + `gen_binop` for the getter).
- `||=` / `&&=` lowered to `target || (target = value)`, duplicating `target` in the AST.

Both re-ran the receiver / index expression.

## Change

Handled entirely in **bytecodegen** by reusing `eval_lvalue`'s single evaluation of the receiver / index (they already land in temp registers) — no synthetic locals or AST rewriting.

- The `*OrWriteNode` / `*AndWriteNode` shapes now lower to `AssignOp(LOr/LAnd, target, value)` (instead of the `BinOp(LOr, target, MulAssign([target],[value]))` wrapper), so `||=` / `&&=` share the same single-evaluation path as arithmetic op-assign.
- A dedicated path handles attribute / index targets: evaluate the receiver / index once, read the current value through the matching getter (from the already-evaluated temps), combine (arithmetic via `BinOpK`; `||=` / `&&=` via a short-circuit branch whose setter fires only when needed), and write back through the setter.
- Side-effect-free targets (local, `self`, ivar, cvar, gvar, constant, dynamic var) keep the simpler re-evaluating path and preserve short-circuit store semantics (e.g. `A ||= 1` still avoids reassigning an already-initialized constant).

## Spec impact

`language/optional_assignments_spec.rb`: **12 failures + 4 errors → 4 + 2** (10 fixed). Remaining failures are the constant-path (`Foo::Bar ||= …`) module-part side-effect cases, tracked separately.

## Tests

- New `op_assign_evaluates_receiver_once` and `op_assign_target_kinds` tests in `tests/arith.rs` (attr `+=` / `||=` / `&&=`, index `+=` / `||=` / `&&=`, and the side-effect-free target kinds with short-circuit and value semantics), all CRuby-verified.
- Full `monoruby` lib suite (1800), `arith` (64), and `method_call` (98) pass on x86-64. The change lives entirely in arch-neutral bytecodegen. No regression in `variables` / `assignments` / `send` specs (18 → 15 failures).

---

### DIFF
--- a/monoruby/src/bytecodegen/expression.rs
+++ b/monoruby/src/bytecodegen/expression.rs
@@ -491,6 +491,31 @@ impl<'a> BytecodeGen<'a> {
                     }
                     return self.gen_safe_nav_op_assign(receiver, op, lhs, rhs, use_mode, loc);
                 }
+                // Attribute (`recv.attr op= rhs`) and index (`base[idx] op=
+                // rhs`) targets have a receiver / index expression that may
+                // carry side effects, so it must be evaluated exactly once and
+                // reused for both the getter and the setter. Every other target
+                // (const / ivar / cvar / gvar / dynamic var) is side-effect-free
+                // to re-read, so the simpler re-evaluating path stays correct.
+                if matches!(&lhs.kind,
+                    NodeKind::MethodCall { .. } | NodeKind::Index { .. })
+                {
+                    return self.gen_op_assign_with_receiver(op, lhs, rhs, use_mode, loc);
+                }
+                match op {
+                    BinOp::LOr | BinOp::LAnd => {
+                        // `target ||= rhs` / `target &&= rhs`: short-circuit so
+                        // the store only fires when needed. Re-reading a
+                        // side-effect-free target is fine, so lower to the
+                        // ordinary short-circuit binop over an inline assign.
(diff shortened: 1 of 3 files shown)


sisshiki1969:
ご要望どおり、synthetic local を使わず bytecodegen 側の temp レジスタで実装し直しました（パーサでの AST 書き換えは廃止）。`eval_lvalue` がレシーバ／インデックスを 1 度だけ temp に評価する点を利用し、その temp から getter を生成 → 演算 → setter で書き戻します。`||=` / `&&=` も `AssignOp(LOr/LAnd, ...)` へ経路統一して同じ 1 回評価を共有します。定数・ivar 等の副作用のない対象は従来どおりの経路のままです。force-push で更新済みです。

---

PR #828, opened 2026-07-05
bytecodegen: evaluate nested masgn targets left-to-right before rhs

sisshiki1969:
## Summary

In a multiple assignment, CRuby evaluates every target's receiver / index subexpression **left-to-right before the right-hand side**, then performs the stores. monoruby did this for flat targets (their `eval_lvalue` ran in the first pass, before the rhs), but a **nested destructure group** (`(a.x, foo), bar = rhs`) was deferred raw and only evaluated during the `ExpandArray` + store phase — *after* the rhs.

So this recorded the wrong order:

```ruby
((ScratchPad << :a; o).a, foo), bar = [(ScratchPad << :b; :b)]
ScratchPad.recorded  # monoruby: [:b, :a]   CRuby: [:a, :b]
```

## Change

`gen_mul_assign` now evaluates a nested group's lvalues up front too, via a new `eval_nested_lvalues` (recursing left-to-right), carrying the captured `LvalueKind`s and the group's splat position. Only the `ExpandArray` + store is deferred (`assign_prepared`), since it needs the parent array element in hand. The receiver / index registers stay live below the rhs, exactly like flat leaf targets — so the evaluation order now matches CRuby while the assigned values are unchanged.

`MulLhs::Nested` was changed from holding raw `Vec<Node>` to the pre-evaluated `{ targets: Vec<MulLhs>, rest_pos }`, and the old `gen_destructure` (which interleaved receiver evaluation with the stores) is replaced by the two-phase `eval_nested_lvalues` / `assign_prepared` pair.

## Spec impact

`language/assignments_spec.rb`: **7 failures + 1 error → 1 failure + 1 error** (6 fixed) — the nested / deeply-nested accessor, `#[]=`, and compounded-constant evaluation-order cases.

## Tests

- New `multi_assign_evaluation_order` (flat, nested, deeply-nested accessor and `#[]=` order via a `$s` recorder) and `multi_assign_nested_values` (value correctness for nested destructuring incl. splat and index targets) in `tests/mul_assign.rs`. CRuby-verified.
- Full `monoruby` lib suite (1800), `mul_assign` (7), `method_call` (99), `variables` (13) pass.

---

### DIFF
--- a/monoruby/src/bytecodegen/expression.rs
+++ b/monoruby/src/bytecodegen/expression.rs
@@ -18,7 +18,14 @@ const LITERAL_CHUNK_LEN: usize = 256;
 ///
 enum MulLhs {
     Leaf(LvalueKind),
-    Nested(Vec<Node>),
+    /// A nested destructure group (`(b, c)` in `(a, (b, c)) = ...`). Its
+    /// members' lvalues are already evaluated (receivers / indices captured
+    /// left-to-right, before the rhs); `rest_pos` records the splat slot for
+    /// the group's own `ExpandArray`.
+    Nested {
+        targets: Vec<MulLhs>,
+        rest_pos: Option<u16>,
+    },
 }
 
 impl<'a> BytecodeGen<'a> {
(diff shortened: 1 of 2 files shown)


sisshiki1969:
The failing `darwin` job is pre-existing macOS-runner flakiness, not this change:

- On both runs (`f5efae4`, `0705563`) the **`Test` step started and never completed** — the job was killed ~50 min in (14:04:25→14:55:50, then 14:58:38→15:48:06). `amd64` ran the identical `bin/test` suite in ~14 min and passed both times, and codecov is green.
- The same ~50-min `Test`-step hang happens on **`master` push-runs of already-merged code**, e.g. the #826 merge commit `0541978` (run 28742080633): `amd64` succeeded, `darwin`'s `Test` step started 13:18:01 and was killed at 14:09:36. That code had already passed `darwin` as a PR — same code, different run, opposite result.
- This change is arch-neutral bytecodegen (multiple-assignment runs mostly in the VM tier; a real infinite loop would hang `amd64` too, since the TraceIR/AsmIR front-end is shared). I built the AArch64 target locally and ran the masgn tests under qemu — all 7 (incl. the two new `multi_assign_*` cases) pass in ~8.6s with no hang.

I couldn't re-run just the `darwin` job (the integration lacks `rerun-failed-jobs` permission — 403), so `0705563` is an empty re-trigger commit. Happy to re-trigger again, but the hang looks independent of the diff. `amd64` + codecov are green.

---

PR #832, opened 2026-07-06
bytecodegen: warn on duplicated literal keys in a hash literal

sisshiki1969:
## Summary

CRuby warns `key ... is duplicated and overwritten on line N` when a hash literal repeats a **literal** key (`{a: 1, a: 2}`, `{1.0 => x, 1.0 => y}`), and routes it through the redirectable `$stderr` — so `-> { eval "..." }.should complain(...)` captures it. monoruby emitted nothing.

## Change

Bytecodegen has no `Executor`, so it can't invoke `$stderr.write` directly. So:

1. During `gen_hash`, detect duplicated **literal** keys (symbol / string / integer / bignum / float / true / false / nil). Runtime-computed keys (`{k => 1, k => 2}`) are left alone, matching CRuby's compile-time check.
2. Buffer the formatted messages on a new `Store::compile_warnings`.
3. Flush them through `$stderr` via a new `Executor::flush_compile_warnings`, called right after compilation and before the compiled code runs — at every compile→run entry point (top-level / `require` via `exec_script`, plus `Kernel#eval` / `instance_eval` / `module_eval` / `Binding#eval`).

Silenced at `-W0`. The float key is rendered with `ruby_float_to_s` so `1.0` prints as `1.0` (not `1`).

This adds **no bytecode instruction** (the compiled bytecode is byte-for-byte unchanged; only a compile-time detection pass + a runtime `$stderr.write` via existing machinery), so both the x86-64 and AArch64 backends are unaffected.

## Spec impact

`language/hash_spec.rb`: fixes "checks duplicated keys on initialization" and "checks duplicated float keys on initialization" (7 → 5 failures). The remaining failures are the `**obj` splat-vs-explicit-key ordering/precedence cases, tracked separately.

## Tests

- New `hash_literal_duplicated_key_warning` in `builtins/hash.rs`: captures `$stderr` via `StringIO` and matches the message for symbol / string / integer / float dup keys, confirms a runtime key does **not** warn, confirms no-dup is silent, and that the overwrite still applies. CRuby-verified.
- Full `monoruby` lib suite (1804) passes.

---

### DIFF
--- a/monoruby/src/builtins/binding.rs
+++ b/monoruby/src/builtins/binding.rs
@@ -124,6 +124,7 @@ fn eval(vm: &mut Executor, globals: &mut Globals, lfp: Lfp, pc: BytecodePtr) -> 
     // guarantees `lfp.self_val()` matches `BINDING_CLASS`.
     let binding = Binding::try_new(lfp.self_val()).expect("self is Binding");
     globals.compile_script_binding(expr, fname, binding, lineno)?;
+    vm.flush_compile_warnings(globals);
     vm.invoke_binding(globals, binding.binding().unwrap())
 }
 
--- a/monoruby/src/builtins/hash.rs
+++ b/monoruby/src/builtins/hash.rs
@@ -4103,6 +4103,37 @@ mod tests {
     }
 
     #[test]
+    fn hash_literal_duplicated_key_warning() {
+        // A hash literal that repeats a *literal* key warns "key ... is
+        // duplicated and overwritten" through `$stderr` (so a redirected
+        // `$stderr` captures it, as CRuby's compile-time warning does).
+        // Only literal keys are checked; a runtime key (`{k => 1, k => 2}`)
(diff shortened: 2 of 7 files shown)


sisshiki1969:
The failing `darwin` job is the known macOS-runner flakiness, not this change: the `Test` step started 01:45:45 and never completed — the job was killed ~49 min in (02:34:10). `amd64` ran the same suite in ~14 min and passed, and `codecov/patch` is green (98.70%). This is the same ~50-min `Test`-step hang that also strikes `master` push-runs on already-merged code (e.g. #828's merge commit, diagnosed [here](https://github.com/sisshiki1969/monoruby/pull/828#issuecomment-4886638402)).

This PR adds **no bytecode instruction** — it's a compile-time duplicate-key check that buffers a message and flushes it through `$stderr` via existing runtime machinery — so the AArch64 codegen is byte-for-byte unchanged. `amd64` + `codecov/patch` are green.

---

PR #882, opened 2026-07-11
Constant lookup recovers the runtime cref in singleton-class bodies

sisshiki1969:
## Summary

Iteration 23 of the ruby/spec language-category improvements. Fixes constant resolution for methods defined inside `class << obj` bodies, clearing the remaining 3 failures in `language/constants_spec.rb` (language category overall: **10 → 7**).

## Problem

```ruby
2.times do |i|
  obj = Object.new
  class << obj
    CONST = i + 1
    def foo; CONST; end
  end
  p obj.foo   # CRuby: 1, 2 / monoruby (before): 2, 2
end
```

Each execution of `class << obj` creates a distinct singleton class, but the `def`s in the body share their iseq across executions. `enter_classdef` re-stamps the shared iseq's static `lexical_context` per execution (last write wins), so methods defined by earlier executions resolved constants against a later execution's class. The same applies to a `class X` nested inside `class << obj` (which is likewise a distinct class per execution).

## Fix

- **`Executor::runtime_innermost_cref`**: when the statically stamped innermost lexical entry is not on the receiver's ancestor chain, substitute the ancestor that owns the executing method (the defining class), recovering the runtime cref. Applied in `search_lexical_stack`, `probe_lexical_stack`, and the superclass-fallback lexical class. A metaclass owner is mapped back to its attached class, so `def self.f` in a class nested under `class << obj` resolves constants in the class rather than the metaclass.
- **`ISeqInfo::in_singleton_lexical`** (new, transitive flag): set on the sclass body iseq and propagated by bytecodegen onto every nested def/classdef/block. Staleness can only arise on lexical chains passing through `class << obj`, so the substitution is gated on this flag — singleton defs (`def a.meth`, `def M.f`) keep their purely lexical cref.
- **Self-keyed inline constant cache**: `ConstCache` gains a `self_class: Option<ClassId>` key. Constant sites in singleton-lexical iseqs (including the sclass body itself and the `||=` definedness route) always key the cache by self — a class/module receiver by its own id, since its metaclass may be unmaterialized. The VM includes the key in the hit condition; the JIT compares it against its compile-time self class (mapped back through the singleton attachment) and requests recompilation instead of folding a value cached for a different self, and never folds a `None`-keyed entry in such methods (its validity is relative to a re-stampable static cref).

## Tests

- `language/constants_spec.rb`: 100 examples, **0 failures** (was 3 failures)
- language category overall: 10 → **7** failures (remaining: super 2 / ensure 1 / rescue 1 / symbol 1 / predefined 1 / pattern_matching 1)
- full `cargo test` suite green
- New integration tests `sclass_constants_resolve_per_object` and `sclass_constants_cover_lookup_paths`: per-object constants in singleton classes, nested classes, `def self.f`, `defined?` probes, ancestor fallback, `||=` cache hits, body-level reads, stale-stamp calls after re-execution, and three interleaved receivers churning the cache while the JIT compiles (JIT warm-up included via `run_test`)
- Patch coverage 96.3% (75.8% before the second commit; the tests written for coverage exposed two real cache-validity bugs, fixed in that commit)

https://claude.ai/code/session_013XLHRVwyWsn5Pp8zgNcZaK

### DIFF
--- a/monoruby/src/bytecodegen.rs
+++ b/monoruby/src/bytecodegen.rs
@@ -792,6 +792,22 @@ impl<'a> BytecodeGen<'a> {
         self.is_escaping_block() || self.singleton_classdef
     }
 
+    /// The lexical chain of the iseq currently being compiled passes
+    /// through a `class << obj` body. Propagated onto every nested
+    /// def/classdef/block so runtime constant lookup knows when the
+    /// statically stamped cref can be stale.
+    fn in_singleton_lexical(&self) -> bool {
+        self.singleton_classdef || self.iseq().in_singleton_lexical
+    }
+
+    fn propagate_singleton_lexical(&mut self, fid: FuncId) {
+        if self.in_singleton_lexical() {
+            if let Some(iseq) = self.store[fid].is_iseq() {
+                self.store[iseq].in_singleton_lexical = true;
+            }
+        }
+    }
+
     fn add_method(
         &mut self,
         name: Option<IdentId>,
@@ -799,8 +815,11 @@ impl<'a> BytecodeGen<'a> {
         loc: Loc,
     ) -> Result<FuncId> {
         let sourceinfo = self.sourceinfo.clone();
-        self.store
-            .new_iseq_method(name, compile_info, loc, sourceinfo, false)
+        let fid = self
+            .store
+            .new_iseq_method(name, compile_info, loc, sourceinfo, false)?;
+        self.propagate_singleton_lexical(fid);
+        Ok(fid)
     }
 
     fn add_classdef(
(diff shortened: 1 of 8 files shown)


sisshiki1969:
Pushed 4568695 to improve patch coverage (75.8% → 96.3% of diff lines measured locally with `cargo llvm-cov`).

Writing the coverage tests exposed two real defects in the first commit, both fixed here:

- **JIT folded `None`-keyed cache entries for any self class.** A `None` key meant "resolved through the static cref", but that validity is relative to the *current* stamp and silently breaks when a later `class << obj` execution re-stamps it. Singleton-lexical methods now always key the constant cache by self, and the JIT bails on a `None` entry there instead of folding it.
- **Constant reads in the `class << obj` body itself shared one cache entry across executions.** The body's own iseq never carried `in_singleton_lexical`, and an ancestor-resolved read with no intervening definition dodges the const-version bump (e.g. two receivers of different classes reading a constant defined on those classes returned the first receiver's value twice). `new_classdef` now flags the body iseq too, and this also covers the `||=` definedness route.

Supporting changes: a class/module receiver keys the cache by its own id (its metaclass may be unmaterialized, which would alias distinct receivers through `Class`), with the JIT mapping its compile-time self class back through the singleton attachment to compare; `runtime_innermost_cref` maps a metaclass owner back to its attached class so `def self.f` in a class nested under `class << obj` resolves constants in the class rather than the metaclass; the ancestry/plausibility helpers moved onto `Store`, shared by the executor and the JIT.

The extended integration test drives every recovery path with genuinely stale stamps (receivers called *after* a later re-execution), the `defined?` probe route, the ancestor fallback, `||=` cache hits, body-level reads, and three interleaved receivers churning the cache while the JIT compiles.

Verified: `language/constants_spec.rb` still 100 examples / 0 failures, language category still 7 remaining, full `cargo test` suite green. The few diff lines still unhit locally are the generic JIT constant-fold path (exercised broadly by the full CI suite) and four lines whose execution is proven by test results but misattributed by lcov on a multi-line method-call chain.

---

PR #946, opened 2026-07-16
Fix core Array/Hash/Enumerable ruby/spec failures

sisshiki1969:
## Summary

Investigated and fixed remaining `core/{array,hash,enumerable}` ruby/spec failures. Across the four categories the unresolved count drops **68 → 34** (34 fixed, no regressions).

| Category | Before (fail/err) | After (fail/err) | Fixed |
|---|---|---|---|
| array | 19 / 26 | 14 / 4 | **27** |
| hash | 12 / 3 | 9 / 2 | **4** |
| enumerable | 6 / 0 | 3 / 0 | **3** |
| regexp | 0 / 2 | 0 / 2 | 0 (deferred) |

## Changes (one commit each)

1. **`Array#rfind` + `find`/`detect` no-block Enumerator** (−19) — implement the Ruby 4.0 `rfind` (reverse scan, `ifnone` proc, tolerates the array shrinking mid-iteration); `find`/`detect` now return a nil-size Enumerator capturing the `ifnone` arg instead of raising `LocalJumpError`.
2. **`Array#sum` Kahan summation + growth tolerance; `Enumerable#inject` too** (−3) — switch to Kahan-Babuška compensated summation once a Float appears (with CRuby's Inf/NaN handling); both block forms re-read the length each step so elements appended mid-iteration are folded in.
3. **`slice_before`/`slice_after`/`slice_when`/`chunk_while` return nil-size Enumerator** (−2) — these chunkers now report `size == nil` (were returning an Array#each enumerator with a concrete size).
4. **`Array#fill` argument handling; `Hash.new` `capacity:`/unknown-kwarg validation** (−6) — fill: reject a 3rd positional arg in block form, treat a nil start as 0, never extend on a fill-to-end, and raise `TypeError` for a Range with a length argument. `Hash.new` accepts the Ruby 4.0 `capacity:` keyword (ignored) and rejects other keywords with `ArgumentError`, via a keyword-aware registration (`FuncInfo::set_effect` keeps `CAPTURE`).
5. **`Array#sort` Integer block-result sign; `Array#[]=` range errors** (−3) — sort derives the ordering from an Integer (Fixnum/Bignum) result's sign without calling `<=>` (matches `rb_cmpint`); `[]=` with a Range raises `RangeError` (not `IndexError`) for an out-of-bounds start and treats an end before the start as a zero-length insertion.
6. **`Hash#transform_keys!` conflict/break semantics** (−2) — reimplement using CRuby's two-phase algorithm so transformed keys don't clobber not-yet-processed originals, and a `break` from the block stops early with the processed pairs applied.

## Testing

- `mspec run core/{array,hash,enumerable,regexp}` — failure/error counts above.
- Rust unit tests for the touched modules pass (`builtins::array::tests` 76, `builtins::hash::tests` 100).

## Deferred (documented, not in this PR)

- **`Regexp.timeout` / `Regexp::TimeoutError`** — requires onigmo match-interruption the Rust binding doesn't expose; a `TimeoutError` that never fires would be a misleading half-API.
- **Warning source-location prefix** (`initialize`/`inject` "given block not used") — blocked on a `caller`/`caller_locations` bug (per-frame line reports the definition line, not the call site) plus native-frame skipping; a broad VM change with backtrace-wide risk.
- **Hash `#delete`/`#shift` during iteration, `Enumerable#map` arity propagation / `grep_v` `$~`, `Array#pack` gaps, recursive-structure hash values, `shuffle!` srand parity** — deeper rubymap/kwarg/svar work with low per-spec payoff.

https://claude.ai/code/session_013XLHRVwyWsn5Pp8zgNcZaK

---
_Generated by [Claude Code](https://claude.ai/code/session_013XLHRVwyWsn5Pp8zgNcZaK)_

### DIFF
--- a/monoruby/builtins/builtins.rb
+++ b/monoruby/builtins/builtins.rb
@@ -108,16 +108,25 @@ class Hash
   def transform_keys!(hash = nil, &block)
     return to_enum(:transform_keys!) { size } unless block || hash
     raise FrozenError.new("can't modify frozen Hash: #{inspect}", receiver: self) if frozen?
-    if hash
-      keys.each do |k|
-        if hash.key?(k)
-          self[hash[k]] = delete(k)
+    # Mirror CRuby's algorithm: iterate a snapshot of the original pairs,
+    # tracking the keys produced so far in `new_keys`. An original key is
+    # removed only when it was not itself produced as a new key by an earlier
+    # step, so transformed keys never clobber not-yet-processed originals
+    # (e.g. `{a:1,b:2}.transform_keys!(&:succ)` => `{b:1,c:2}`). A `break`
+    # from the block stops early, leaving the pairs processed so far applied.
+    new_keys = {}
+    to_a.each do |key, val|
+      new_key =
+        if hash && hash.key?(key)
+          hash[key]
         elsif block
-          self[block.call(k)] = delete(k)
+          block.call(key)
+        else
+          key
         end
-      end
-    else
-      keys.each { |k| self[block.call(k)] = delete(k) }
+      delete(key) unless new_keys.key?(key)
+      self[new_key] = val
+      new_keys[new_key] = nil
     end
     self
   end
--- a/monoruby/builtins/enumerable.rb
+++ b/monoruby/builtins/enumerable.rb
@@ -478,7 +478,8 @@ module Enumerable
       end
     end
     result << current unless first
-    result.each
+    # Returns an Enumerator of unknown size, so `enum.size` is nil (CRuby).
+    Enumerator.new { |__yielder| result.each { |__chunk| __yielder << __chunk } }
   end
 
   def slice_when(&block)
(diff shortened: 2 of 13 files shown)


sisshiki1969:
Closing as superseded. Every change here has already been merged into `master` independently, so this PR is fully redundant:

- Array#rfind + find/detect no-block enumerator → #923
- Array#sum Kahan + growth tolerance → #929
- Array#fill / Hash.new(capacity:) / slice_before size → #925
- Array#sort sign + Array#[]= range semantics → #934 (+ #932)
- Hash#transform_keys! conflict/break → #928
- Enumerable#inject growth tolerance → #940
- base runtime commits (super semantics, eval-site def) → #864

`master` also already covers the items this PR deferred (#939 warning caller-location, #938 Hash#map arity, #926 pack). No unique work remains here, so closing.

---

PR #1087, opened 2026-08-10
Hash: add __key_at / __value_at index intrinsics, inlined to a C-ABI call

sisshiki1969:
## Summary

Ruby-level builtins iterate with `each { ... yield ... }`, which costs **two block entries per element**: the block handed to `each`, then the user's block. Positional access lets that be a `while` loop over indices instead, leaving just the user's block — **one entry per element**.

```ruby
def to_h
  h = {}; i = 0; n = size
  while i < n
    pair = yield __key_at(i), __value_at(i)
    h[pair[0]] = pair[1]
    i += 1
  end
  h
end
```

`HashRef::entry_at` is O(1) for all three representations (the inline pairs and `RubyMap`'s entry vector are both position-addressable).

## Inlining

Both methods are registered with `define_builtin_inline_func`, so a JIT'd call site becomes a **direct C-ABI call** instead of going through method dispatch and the builtin trampoline.

The call sequence is the one `Hash#[]` already uses: `emit_hash_index` takes the target address as a parameter, so **both backends pick this up with no new arch-specific code** — the diff is two Rust files, nothing under `codegen/arch/`.

## Measurements (release build, `to_h`-shaped workload, ns/pair)

| | n=1,000 | n=100,000 |
|---|---|---|
| `each` + `yield` (current) | 177.1 / 185.6 | 312.5 / 294.8 |
| `while` + intrinsic | **138.7 / 127.4** | **250.6 / 232.4** |
| speedup | **~1.4x** | **~1.25x** |

No allocation either — contrast a `keys`-based Ruby `each`, which pays an Array plus N lookups and came out *slower* than the current code (500 vs 330 ns/pair at n=100000).

## What this does not reach, and why

A Rust `Hash#to_h` still beats both (~109 / ~199), because two intrinsic calls per pair remain. A per-element builtin call measured **13.3 ns** against **9.2–10.5 ns** for a block entry, so the win here comes from *halving the entries*, not from the call being cheap.

Closing the rest means lowering these to machine code. That needs the vendored `rubymap`'s `Bucket` / `IndexMapCore` pinned with `#[repr(C)]` (plus `offset_of!` assertions) before the JIT can hardcode offsets — Rust guarantees no layout today, so hardcoding would be unsound across compiler versions. Deliberately not attempted here.

For reference, an inlined-to-machine-code index (`Array#[]`) measures **3.2 ns**, so the headroom is real.

## Naming and testing

The index is a position in the hash's own entry order, not part of the public Ruby interface — hence the `__` prefix.

Since CRuby has no such methods, the test computes the same value via `to_a` when they are absent. CRuby therefore checks our ordering, rather than us asserting it against ourselves. Covered: mixed key types, the large (`Map`) representation, `compare_by_identity` (`IdentMap`), an empty hash, agreement with `each`'s order, and `nil` for out-of-range and negative indices.

Full suite: **3073/3073**.

https://claude.ai/code/session_011nsgwMopV95G3pfLbLMAPA

---
_Generated by [Claude Code](https://claude.ai/code/session_011nsgwMopV95G3pfLbLMAPA)_

### DIFF
--- a/monoruby/src/builtins/hash.rs
+++ b/monoruby/src/builtins/hash.rs
@@ -45,6 +45,24 @@ pub(super) fn init(globals: &mut Globals) {
     globals.define_builtin_func(HASH_CLASS, ">", gt, 1);
     globals.define_builtin_func(HASH_CLASS, ">=", ge, 1);
     globals.define_builtin_inline_func(HASH_CLASS, "[]", index, inline_gen2!(hash_index), 1);
+    // Positional entry access, used by Ruby-level iteration written as a
+    // `while` loop over indices instead of an `each` block. Internal: the
+    // index is a position in the hash's own entry order, which is not part of
+    // the public Ruby interface.
+    globals.define_builtin_inline_func(
+        HASH_CLASS,
+        "__key_at",
+        key_at,
+        inline_gen2!(hash_key_at),
+        1,
+    );
+    globals.define_builtin_inline_func(
+        HASH_CLASS,
+        "__value_at",
+        value_at,
+        inline_gen2!(hash_value_at),
+        1,
+    );
     globals.define_builtin_func(HASH_CLASS, "[]=", index_assign, 2);
     globals.define_builtin_func(HASH_CLASS, "clear", clear, 0);
     globals.define_builtin_func(HASH_CLASS, "replace", replace, 1);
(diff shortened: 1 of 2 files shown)


sisshiki1969:
Superseded by #1086, which now carries both `Hash` commits in one PR (this branch's commit was cherry-picked onto `claude/hash-to-h-rust`). Closing in favour of that one — no work is lost.

---

PR #1156, opened 2026-08-20
JIT: heal monomorphic-compiled sends that turn polymorphic — −16% on activerecord

sisshiki1969:
Stage 1 of the despecialization work (the ~23% chain-deopt share on activerecord). A general send compiled while its PMC held one receiver class kept a single-class guard with a **plain** deopt: once a second class arrived, the site deopted on every off-class receiver for the rest of the run. The polymorphic lowerings (class-set guard, PIC) and the counter-gated recompile machinery all existed — nothing connected a receiver-guard miss to them outside BinCmp's Part B.

Measured on activerecord before this change: `_read_attribute`'s `fetch_value` site alone failed its class guard **13,206** times/run (AttributeSet vs LazyAttributeSet), `deserialize` 7,440, `key?` 3,100, plus `nil?`/`platform`/`create` — all "compiled monomorphic before the variance arrived, never re-learned."

## What it does

- **`RecvMissMode::{Plain, PartB, Learn}`** replaces `compile_method_call`'s `recompile_on_recv_miss` bool. General sends pass `Learn`: while the PMC is still monomorphic, the receiver guard exits through a counter-gated `BecamePolymorphic` recompile — each deopt's VM re-execution feeds the PMC (`find_method`'s slow path), so the fresh compile sees the variance and builds the class-set guard or PIC. Once the PMC shows ≥2 classes, those gates were already tried *at this compile* and declined, so the guard plain-deopts — a **ratchet**: no recompile-per-N-misses livelock. BinCmp keeps its Part B semantics unchanged.
- **`RecompileTarget::{Whole, Specialized}`** replaces the `Option<BytecodePtr>` position in the `RecompileDeoptimize` side exit, lifting the "specialized JIT recompiles via an idx, not a position" exclusion: a specialized (inlined) body's receiver guard now heals through `jit_recompile_specialized` (`COUNT_DEOPT_RECOMPILE_SPECIALIZED`, re-pointing the caller's patch point) — the same route its class-version guard already takes, and it passes straight through #1155's salvage (which is reason-gated on `ClassVersionGuardFailed`). Both arch backends lower the new target.
- **Block ROOTs stay on the plain deopt**: a whole-method recompile rebuilds a block body under the wrong argument convention (the `guard_const_version` comment records the `caller_locations` breakage), so a `Learn` exit is only emitted where a sound recompile target exists.

## Results

| | before | after |
|---|---|---|
| activerecord (avg of last 10 iters, ×3 runs) | 2,375ms | **2,004ms (−16%)** |
| `_read_attribute` class-guard failures | 13,206 | 3,020¹ |
| `_has_attribute?` `key?` / `nil?` / `platform` / `create`→`new` | top-20 deopt sites | gone |
| fib / mandelbrot / binarytrees / optcarrot | — | unchanged² |
| full suite | | **3,494 pass** |
| full suite under `gc-stress` | | **3,495 pass** |
| activerecord 14-point output vs CRuby 4.0.2 | | byte-identical |

¹ The residual is the learn counter itself: ~60 specialized instances × up to 50 misses each before their one-shot heal.
² Monomorphic sites emit the identical guard shape (the `Learn` exit only changes where the guard *jumps*, and only until the site is known-mono forever).

New tests: `pic_late_polymorphism_heals` / `pic_late_polymorphism_heals_specialized` — a site that goes polymorphic only after the JIT compiled it, in a root and in an inlined body.

**Known residual / follow-up**: the `deserialize` site (7.4k deopts) lives in `LazyAttributeSet#fetch_value`'s **block**, compiled as a block ROOT — healing it needs a block-convention-preserving whole recompile. Stage 2 (argument-type generalization: the 150k `Array#initialize`/`Array#map` deopts) is also separate.

---

### DIFF
--- a/monoruby/src/codegen/arch/aarch64/compile/mod.rs
+++ b/monoruby/src/codegen/arch/aarch64/compile/mod.rs
@@ -207,15 +207,12 @@ impl Codegen {
                         label
                     }
                 }
-                // Treat recompile-deopt as a plain deopt for now: fall back to
-                // the interpreter without the counter-gated recompile (still
-                // correct, just not yet self-optimizing).
-                SideExit::RecompileDeoptimize(pc, wb, reason, position, chain) => {
+                SideExit::RecompileDeoptimize(pc, wb, reason, target, chain) => {
                     let label = self.jit.label();
                     lir.push(LInst::SideExit {
                         kind: LSideExitKind::RecompileDeopt {
                             reason,
-                            position,
+                            target,
                             chain,
                         },
                         pc,
@@ -1350,13 +1347,13 @@ impl Codegen {
                 // `error_body` if the recompile itself raised a FatalError.
                 LSideExitKind::RecompileDeopt {
                     reason,
-                    position,
+                    target,
                     chain,
                 } => {
                     let deopt_body = self.jit.label();
                     let error_body = self.jit.label();
                     self.jit.bind_label(entry);
-                    self.emit_recompile_deopt(position, &deopt_body, Some(&error_body), reason);
+                    self.emit_recompile_deopt(target, &deopt_body, Some(&error_body), reason);
                     self.a64_gen_deopt(pc, &wb, deopt_body, loop_jit_spill_bytes, base, chain);
                     self.a64_gen_handle_error(pc, &wb, error_body, loop_jit_spill_bytes, base, chain);
                 }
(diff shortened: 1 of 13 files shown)


sisshiki1969:
Note on the codecov/project −0.01%: the 2 uncovered lines are the `RecompileTarget::Specialized` arm of x86's `emit_recompile_deopt`. On x86 that arm is structurally unreachable today — specialized recompile side exits are lowered in `side_exit_with_label` (which calls `gen_recompile_specialized` directly), and the main-body `LInst::RecompileDeopt` this function serves is always a `Whole` target. The arm is kept so the shared `compile_shared` call site type-checks identically on both arches (aarch64 routes its side exits through the same function, where the arm *is* exercised — the darwin job covers it). amd64/darwin and codecov/patch (98.2%) are green.

---

PR #1168, opened 2026-08-21
Heal block-root constant guards; separate the measurement feature tiers

sisshiki1969:
Three findings from re-measuring the activerecord benchmark after #1164/#1165. The first is the substantive one; the other two came out of the measurements themselves.

## 1. A block root can now heal its constant-version guard

A block-style root cannot take the salvaging *recompile* entry — that entry recompiles whatever `lfp.func_id()` names, and re-entering the compiler as if a block frame were a method rebuilds the body under the wrong argument convention. So block roots took a bare deopt instead, and a bare deopt never heals: the global constant version does not move back, so a single `X = ...` anywhere left every block that had folded a constant failing its guard **on every later call, for the rest of the process**.

On the activerecord benchmark that was **69% of all constant-version deopts** — 59,278 of 85,818 over two iterations — concentrated in a handful of blocks:

```
31,400  block in ActiveRecord::Result#indexed_rows
18,137  block (2 levels) in Gem::BasicSpecification#have_file?
10,840  block in Gem::BasicSpecification#have_file?
 5,919  block in Gem::Requirement#satisfied_by?
 4,000  block in ActiveRecord::PredicateBuilder::ArrayHandler#call
```

It also explains a gap that had looked like a counting error: 88,286 constant-version deopts against only 6,798 salvage attempts. The missing ~81,000 were the block roots — they called nothing at all. And the counters made the permanence visible: at `FIXED_ITERS=3` and `FIXED_ITERS=6` every constant-version counter was *identical* (`const_version incs: 2133`, `recovery attempts: 6798`, `salvaged: 3775`), i.e. all constant activity happens during warmup and none in the steady loop — so nothing was ever going to move the version back.

The objection to recompiling does not reach salvage: it re-reads the unit's recorded folds, compares them against the live constants and writes one word, building no frame. `GuardConstVersion` now carries a `ConstMiss` saying what a miss tries before deopting — the recompile entry as before, or `Salvage`. Both backends share the salvage entry, since it emits no code.

## 2. Measurement features: counter tier vs trace tier

`profile` pulled in `dump-bc`, which it never used, and the two chain-deopt trace lines were gated on `deopt`/`profile`. Chain escalation is a per-side-exit runtime operation, not a per-deopt event — hundreds of thousands of calls per iteration — so logging it wrote 160 MB of stderr in five iterations. Measured ladder (40 iterations, one session):

```
(none)              437.6 ms/iter      1 KB
jit-log             469.1 ms/iter    1.5 MB   +7%
profile             481.9 ms/iter     22 KB   +10%
chain-deopt-log    3269.4 ms/iter    265 MB   7.5x
deopt              6118.0 ms/iter    504 MB    14x
```

A `deopt` build was 14x slower than release, which hides the costs such a build exists to show. The trace lines move to an opt-in `chain-deopt-log`.

## 3. An unconvergeable recompile loop on an `Integer` receiver guard

`guard_class` for `Integer` is a Fixnum-immediate tag test, so a heap Integer (BigInt) has the same `ClassId` and fails it. The PMC counts classes and cannot see that split, so it stays at one entry, the Learn ratchet never engages, and every recompile re-emits the identical test.

`ActiveModel::Type::Integer#out_of_range?` is the case in the wild — it compares against `1 << 63` for an 8-byte column, always a BigInt, so its guard can never pass. It recompiled **8,756 times over 40 iterations**, 636 ms of the 986 ms spent in the JIT. Removing the loop cut JIT compile time to 350 ms; wall clock did not move outside noise (3% of the run, below this environment's ±5%, measured interleaved with a YJIT control).

## What was verified, and what was not

Verified here: x86-64 and aarch64 both compile clean across default / `deopt` / `profile` / `jit-log` / `emit-asm`; the `redefine` suite passes (24 tests, including two new ones that drive the salvage entry on a block frame, one of them nested so the frame's `self` comes from an outer block).

**Not verified here:** that a block root now actually heals. The container was recreated between the investigation and this implementation — the ruby-bench checkout, the activerecord gems and the probe scripts are gone, and the host Ruby is 3.3.6 rather than the pinned 4.0.2, so the benchmark cannot be re-run and the full test suite hangs spawning the wrong Ruby. Every number above is from measurements taken before that, and I have re-quoted rather than re-produced them.

I tried to substitute a micro-repro and could not build one: in every shape I tried the block was inlined into its enclosing method and ran through the specialized route, which already heals via its `idx`. The block-root body compiles but is not what executes. The failing configuration needs a hot block whose root body is the live one, as `ActiveRecord::Result#indexed_rows`'s block is — I could not arrange that in isolation, and I would rather say so than present the two passing tests as evidence they are not.

The new oracle entries the added tests generate were deliberately left out of this diff, since this container's Ruby is 3.3.6; CI will record them against 4.0.2.

---

### DIFF
--- a/monoruby/Cargo.toml
+++ b/monoruby/Cargo.toml
@@ -40,7 +40,16 @@ gc-verify = []
 # code paths (LoadSpill / StoreSpill, spill-aware AsmInst lowerings,
 # 16-byte spill-region alignment).
 stress-spill-pool = []
-profile = ["dump-bc", "dump-traceir"]
+# Counters only: `profile` needs `TraceIr::format` for its deopt-site table,
+# not the bytecode dumps `dump-bc` brings.
+profile = ["dump-traceir"]
+# Trace every chain-deopt escalation to stderr. Opt-in, and deliberately not
+# implied by `deopt` / `profile`: escalation is a per-side-exit runtime
+# operation rather than a per-deopt event, and it fires hundreds of thousands
+# of times per activerecord iteration. Logging it unconditionally in `deopt`
+# builds wrote 160 MB of stderr in five iterations and made those builds 7x
+# slower than release, hiding the very costs a measurement build should show.
+chain-deopt-log = []
 perf = []
 # §5 Layer-② extraction: stage-1 *placement shadow*. Records, in emission order,
 # every physical FP placement that ③ commits (one entry per `PhysMap::resolve`),
--- a/monoruby/src/codegen.rs
+++ b/monoruby/src/codegen.rs
@@ -1691,7 +1691,7 @@ impl Codegen {
             if !self.check_vm_address(ret)
                 && let Some(replay) = self.chain_deopt_table.get(&ret).cloned()
             {
-                #[cfg(any(feature = "deopt", feature = "profile"))]
+                #[cfg(feature = "chain-deopt-log")]
                 eprintln!("### chain deopt: frame return {ret:?} -> chain conversion");
                 plan.push(ChainConversion::new(cfp, prev_cfp, replay, stub));
             }
--- a/monoruby/src/codegen/arch/aarch64/compile/method_call.rs
+++ b/monoruby/src/codegen/arch/aarch64/compile/method_call.rs
@@ -1,5 +1,6 @@
 use super::*;
 use monoasm_macro::monoasm_arm64;
+use crate::codegen::jitgen::lir::ConstMiss;
 
 // ---- Object#send inline symbol cache (aarch64 twin of the x86_64 copy in
 // arch/x86_64/compile/method_call.rs; the two files are arch-selected so the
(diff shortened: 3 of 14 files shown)


sisshiki1969:
Closing the loop on the verification this PR said it could not do: with the environment rebuilt (Ruby 4.0.2+YJIT from source, AR 8.1.3.1, ruby-bench stock activerecord), the healing is confirmed empirically. Steady-state const-version deopts over 2 fixed iterations, same machine, before (`0ff506d`) vs after (this PR):

```
                                  before    after
const-version deopts (steady)     24,992      110   (-99.6%)
  of which block roots            24,887       10
  block in Result#indexed_rows    18,900        0
  block in ArrayHandler#call       4,000        2
const-salvage attempts (run)       6,801    9,463   (block roots now reach the entry)
```

The residual 110 is the designed shape — one deopt per unit per version move, then the salvaged guard passes. The `cache stale` Defer path leaves no residue: the deopted execution refreshes the VM cache and the retry salvages.

---

PR #1181, opened 2026-08-25
Run the ensure on a non-local exit out of a specialized chain

### LINKED ISSUE #1179: JIT divergence: an ensure run by `break` writes an outer local the loop unit then reads stale
## Summary

When a `break` unwinds through an `ensure` clause that writes an **outer** local (a dynvar of the enclosing loop-JIT frame), the write is performed by the VM — but the loop-JIT unit keeps reading its own stale view of that local for all remaining iterations.

Found while building the regression shapes for #1178; it is **not** the sp-undo defect fixed there (evidence below).

## Reproduction

```ruby
def g
  a = 0.0; c = 2.0
  i = 0
  while i < 300
    [1].each do |x|
      begin
        a += c
        break if i == 200
      ensure
        c += 0.5
      end
    end
    i += 1
  end
  a
end
p g
```

| | result |
|---|---|
| CRuby (4.0.5) | `23025.0` |
| monoruby `--no-jit` | `23025.0` |
| monoruby JIT, default pool | **`22975.5`** |
| monoruby JIT, `stress-spill-pool` | **`22975.5`** |

Verified on `master` (`225c79cd`) and unchanged by the #1178 fix. arm64-apple-darwin native; the mechanism is arch-neutral as far as we can tell (not verified on x86-64 yet).

## Analysis

The deficit is exactly `0.5 × 99` — one missed `c += 0.5` per iteration for the 99 iterations *after* the `break`. That is: the breaking iteration's `ensure` increment (executed by the VM during the non-local unwind) never reaches the view of `c` the loop-JIT unit keeps reading.

The shape matters:

- the `while` loop is an OSR (loop-JIT) unit holding `c` (and `a`) in registers / as its own view;
- the block is inlined into that unit; the `begin`/`ensure` forces the **non-specialized** `BlockBreak` lowering (`iter_caller_specialized_ids` refuses a chain that crosses an exception handler);
- `break` → `err_block_break` → `entry_raise` → `handle_error` finds the `ensure`, and the VM runs `c += 0.5`, writing the *slot*;
- control comes back to the loop, which still trusts its pre-break register/const view of `c`.

Dropping the `break` (so the `ensure` runs only on the normal path) makes JIT output agree with CRuby, which isolates the divergence to the non-local-exit path.

Same family as the block-passing-call float staleness fixed in #1172/#1173 — a call that can write an outer local must invalidate the frame's kept view — but through the non-local exit rather than the plain call return. The suspended loop frame is presumably converted / resumed without the "forget what the block may have written" step that the ordinary call path performs.

## Regression record

`monoruby/tests/loop_jit_exit_sp.rs::break_running_an_ensure_that_writes_an_outer_local` (added in #1178) carries this exact shape as an `#[ignore]`d test with the analysis in its doc comment. When this issue is fixed, removing the `#[ignore]` turns it into the regression test.

sisshiki1969:
Fixes #1179 — and corrects its analysis.

## The reported theory was wrong

The issue blamed a stale view: the `break`'s `ensure` writes an outer local in the VM, and the loop unit keeps reading the old value. Counting the ensure executions through a global disproved that — **the ensure never ran**:

```ruby
[1].each do |x|
  begin
    break if true
  ensure
    $n += 1        # 99 of 300 executions — exactly the pre-OSR, interpreted ones
  end
end
```

`handle_error` (instrumented) was never entered for a post-OSR `break`. The integer variant reproduced with an identical deficit, ruling the FPR file out entirely.

## Two defects, one path

**1. The specialized teardown never asks whether the exit itself sits in a handler region.**

`iter_caller_specialized_ids` / `method_caller_specialized_ids` check only the *suspended* frames of the chain (by call-site pc, `check_exception_handler`). Nothing tested the `break` / `return` instruction's own pc — so the loop unit lowered the exit to `BlockBreakSpecialized`, a static `lea rbp += Σ; leave; ret` that consults no exception table, and every `break` compiled into the OSR unit skipped its `ensure`.

Fix: `pc_in_handler_region` — a non-local exit inside a begin/ensure refuses the specialized lowering and takes `err_block_break` / `err_method_return` → `handle_error`, which runs the ensure (`defer_unwind`).

**2. Exposed immediately: the non-specialized lowerings don't materialize locals.**

`Raise` / `Retry` / `Redo` all do `locals_to_S` before leaving, because `handle_error` may resume *this frame* in the interpreter at its ensure body — which reads slots. `MethodRet` / `BlockBreak` did not, so a block-local held in a register (or folded as a constant) was nil when the ensure read it: `undefined method '+' for nil`. Fix: the same `locals_to_S`, on the non-specialized arm only — the specialized teardown never resumes the frame.

## Tests

- The #1178 `#[ignore]` divergence record is un-ignored and its doc comment rewritten with the corrected analysis.
- Two sharper shapes join it: the ensure-count-via-global variant (the 99/300 measurement), and the all-integer variant that disproved the stale-view theory.

| | result |
|---|---|
| aarch64 native, `stress-spill-pool` | 3323 passed, 1 skipped |
| aarch64 native, default pool | 3323 passed, 1 skipped |

The change is arch-neutral (TraceIr → AsmIr selection in `jitgen/compile.rs` + `context.rs`); an x86-64 run is needed before merge. x86 has the same two defects structurally — `BlockBreakSpecialized`'s teardown is shared — so the amd64 CI leg is a real check, not a formality.

### DIFF
--- a/monoruby/src/codegen/jitgen/compile.rs
+++ b/monoruby/src/codegen/jitgen/compile.rs
@@ -907,8 +907,8 @@ impl<'a> JitContext<'a> {
             TraceIr::MethodRet(ret) => {
                 state.flush_gp(ir);
                 assert!(state.no_capture_guard());
-                state.load(ir, ret, GP::Rax);
-                if let Some((spec_ids, extra)) = self.method_caller_specialized_ids() {
+                if let Some((spec_ids, extra)) = self.method_caller_specialized_ids(pc) {
+                    state.load(ir, ret, GP::Rax);
                     ir.push(AsmInst::MethodRetSpecialized {
                         rbp_offset: DynVarOffset::Hint {
                             ids: spec_ids,
@@ -916,6 +916,14 @@ impl<'a> JitContext<'a> {
                         },
                     });
                 } else {
+                    // The non-specialized path goes through `handle_error`,
+                    // which may resume *this* frame in the interpreter (an
+                    // ensure body protecting the `return`) — so every local
+                    // must be materialized in its slot first, exactly like
+                    // `Raise`. The specialized teardown above never resumes
+                    // the frame and skips this.
+                    state.locals_to_S(ir);
+                    state.load(ir, ret, GP::Rax);
                     ir.push(AsmInst::MethodRet(pc));
                 }
                 let result = state.as_return(ret);
@@ -925,8 +933,8 @@ impl<'a> JitContext<'a> {
             TraceIr::BlockBreak(ret) => {
                 state.flush_gp(ir);
                 assert!(state.no_capture_guard());
-                state.load(ir, ret, GP::Rax);
-                if let Some((spec_ids, extra)) = self.iter_caller_specialized_ids() {
+                if let Some((spec_ids, extra)) = self.iter_caller_specialized_ids(pc) {
+                    state.load(ir, ret, GP::Rax);
                     ir.push(AsmInst::BlockBreakSpecialized {
                         rbp_offset: DynVarOffset::Hint {
                             ids: spec_ids,
@@ -934,6 +942,10 @@ impl<'a> JitContext<'a> {
                         },
                     });
                 } else {
+                    // Same as `MethodRet`: `handle_error` may resume this
+                    // frame at its ensure body, which reads the slots.
+                    state.locals_to_S(ir);
+                    state.load(ir, ret, GP::Rax);
                     ir.push(AsmInst::BlockBreak(pc));
                 }
                 let result = state.as_return(ret);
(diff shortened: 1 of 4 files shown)


sisshiki1969:
Superseded by #1180, which merged first — the x86-64 session and this one root-caused #1179 independently and arrived at the same two-part fix (the exit-site handler-region test + `locals_to_S` on the generic arm). #1180's version is the better one:

- it verified the `MethodRet` twin with a real reproduction on x86-64 (ensure ran 38/300 times), which this branch could not construct on aarch64;
- its `in_protected_region` keeps the helper signatures unchanged;
- its test file additionally covers the raising-frame-float homing directly and the intermediate-frame contrast case.

Verified #1180's merged master on aarch64 native (arm64-apple-darwin): all nine #1179 repro shapes match CRuby, full suite green at both pool sizes.

The two diagnostic shapes unique to this branch (the ensure-execution count via a global, and the all-integer variant) may follow as a small test-only PR.

PR #1194, opened 2026-08-28
JIT: unify joins — AbstractState spans the whole trace chain

sisshiki1969:
Follow-up to #1193, driving its 「トレース中のフレームのjoinも外側のスコープのjoinも同じようにできるはずです」 observation to its conclusion: joining the frames suspended mid-trace and joining the lexical outer scopes are the same operation, so the ordinary state machinery (clone at a branch, join at a merge, bridge the difference) replaces the bespoke side channels one by one. `doc/trace_chain_joins.md` (added here) is the full design record.

## What changed

**Returns are merge edges (`f511a517`).** A plain `Ret` in a specialized callee records the full chain as an entry edge of the caller's continuation. After the body compiles, the edges' chains are joined; each edge gets an outline segment bridging its state to the join — including the *surrender write* for a kept-`C` claim that this path's subtree gave up (`C ⊔ S → S` emits the chain-addressed literal store) — so keeping a constant across a call is now decided per return path instead of all-or-nothing. The caller resumes from the join, with a deliberate asymmetry: outer levels take the joined chains whole, while the caller level keeps its parked frame's own-timeline bookkeeping (pc, liveness, hints, invariants) and overlays only the join-validated slot claims (kept `C`s, spill-homed `Sf` promotions, float-read marks). Full-chain replacement was tried and broke real code (`Integer#downto`, `Array#permutation`); the bisection that led to the asymmetry is in the doc.

**Nested compiles get the live chain (`0a40cc61`–`c598a9ac`).** Read/keep decisions for outer-frame floats consult the live chain rather than the parked copies, and a nested compile enters with the caller's per-branch chain instead of a frame-entry checkpoint. The no-capture invariant is maintained lexically on the live chain (unset up the lexical links at a capturing call, re-proved up the same links by the capture guard — sound because `branch_if_captured`'s meta check also catches an ancestor `move_frame_to_heap`). The first cut ratcheted outer invariants false and silently degraded chain-addressed loads (+22% on block-read benchmarks, found by instruction-count probes); the lexical pair fixes it and let the parked-invariants restoration be removed.

**Stage-C adoption is ordinary state (`fbde91d2`).** Float-read marks land on the owner frame in the state and travel like any monotone hint (ORed through joins, carried by return chains, merged by the resume overlay). Loop-entry adoption reads them off the back edge's state; the analysis clone exports only vetoes. The adopted `Sf(Float)` binds on the live loop-entry state and its scope is handled by the joins themselves — a path bypassing the head meets it back to `S` — so the revert machinery is deleted.

**No-return-path calls resume from the entry chain (`71f27013`).** A callee that compiles no plain `Ret` leaves the caller's continuation unreachable on compiled paths; its placeholder state is now rebuilt from the call's own entry chain re-widened by the callee-era widen log, retiring the parked copies' fallback-resume role (`adopt_outer` is gone).

**The block-entry fixpoint is deleted (`57791e78`).** `converge_block_entry` probed a yielded block in a throwaway context to discover which outer constants it gives up — an approximation of "block re-entry is a back edge" from before the chain spanned the trace. The chain-wide joins now cover every re-entry path: a yield inside the unit's own loop demotes outer `C`s at that loop's backedge join, straight-line repeated yields thread the widen through the live chain, and re-entry through a loop outside the chain can only be a generic yield, where `forget_constants` already drops every claim. With it go its helpers — the last readers of the parked copies' constant accounting — and one probe compile per specialized yield.

After this, the parked frame copies are no longer a truth channel for cross-frame claims: what remains is the suspended frame's own resume record, the spill-home id space, and the stage-2 promotion gate (allocator engineering, left for a future change).

## Verification

Each commit was verified with the full default and `stress-spill-pool` suites (3575/0 each), aarch64 `cargo check`, and benchmarks flat within the noise band. Adversarial re-entry shapes (callee-loop-driven blocks, two-level nesting writing the outermost `C`, straight-line double yields, caller-side `while`, float constants) run bit-identical vs CRuby, and the no-return-path fallback was probe-verified firing on all-raise / all-break / non-local-return shapes with live outer floats.

---


sisshiki1969:
On the two codecov signals on this head (`5e6a8c9`), for the record:

**`codecov/project` failure (−0.01%)** is run-to-run coverage noise, not a regression from this diff: the previous head (`5529e9c`) measured the *same* non-test code at misses +0 vs base, and this head — which only added tests — measured +6, all outside the diff. JIT coverage varies per run (which paths compile depends on thresholds/timing). CI itself is green on both jobs.

**`codecov/patch` (97.4%)** — the 7 remaining lines in `context.rs` were investigated individually. The earlier codecov report on `5529e9c` did surface a real gap (the no-return-path fallback resume was only probe-verified), which `5e6a8c9` closed with three regression tests, probe-confirmed to take that branch. What's left is:

- four defensive `continue` guards in `adopt_outer_loop_views` (mark at an out-of-chain position, adoption gate with no-capture already lost at the loop head, a frame with no parked state, no frame-offset hint) — two rounds of targeted shapes failed to reach them, consistent with them being belt-and-suspenders rather than live paths;
- the `debug_assert` arm documenting that a suspended frame's bare `F` is path-invariant across return-edge joins.

Covering these would mean removing the guards, which is the wrong trade. Leaving them as is.

---

PR #1234, opened 2026-09-02
Trim the work around a regexp search in match and gsub-with-block

sisshiki1969:
## What

Callgrind on addressable/equality showed the regexp builtins spending about a quarter of their time outside Onigmo. Two of those overheads are removable without touching the engine:

- **`String#match` / `Regexp#match` built their MatchData twice**: once as `$~` inside `captures_from_pos`, then a second object (with a second haystack view) to return. CRuby returns the `$~` object itself, so `match_one` and `rmatch` now hand back `Executor::current_match_data()` — `str.match(re).equal?($~)` holds here too (the block form as well). The fallback that builds the object only covers a frame with no svar container.
- **`gsub` with a block froze a snapshot of the receiver and spliced a result out of it even when nothing matched.** For a normalizing `gsub(/[^allowed]/) { … }` that is nearly every call: 83 of 150k on this benchmark ran the block. It now probes the live receiver first — no block can run before the first match, so nothing can move the buffer under the view — and on a miss returns one plain copy, leaving `$~` cleared as before. On a hit the walk over the frozen snapshot starts at the probe's match position instead of re-scanning from the start (the same non-overlapping walk `captures_iter` does, including the empty-match-after-a-match skip), so a matching call costs no extra search.

## Numbers

| | master | this PR | YJIT |
|---|---|---|---|
| `s.match(/(\w+)\.(\w+)/)` | 613 ns | 419 ns | 564 ns |
| `s =~ /com$/` | 324 ns | 224 ns | 279 ns |
| `gsub(/[^a-z…]/) {}` no match, 11 B | 424 ns | 359 ns | 437 ns |
| `gsub(/[^a-z…]/) {}` no match, 200 B | 3289 ns | 3370 ns | 3611 ns |
| `gsub(/[^a-z…]/) {}` 3 matches, 21 B | 1808 ns | 1750 ns | 2746 ns |

The 200 B case is the Onigmo scan itself (about 16 ns per byte for a negated class, in CRuby too). On addressable/equality the iteration executes 1.75% fewer instructions (6.163 G → 6.055 G); wall-clock is within noise for equality and about 4% better for addressable/normalize.

## What stays

The remaining wrapper cost is the `onig_region_new` / `onig_region_free` pair per search (2.4% of the instructions on this benchmark; CRuby reuses its region). That lives in the onigmo-regex crate — a thread-local free list in `Region::new` / `Drop` removes it — and is a separate change there.

## Tests

- `match_returns_the_saved_backref`: `String#match` / `Regexp#match` identity with `$~` (plain, positional, named-capture, block form, String pattern, no match) against a live CRuby.
- `gsub_block_without_a_match_probes_first`: the no-match copy is unfrozen and independent with `$~` cleared, `$~` inside the block on a hit, a receiver mutated by the block still raises, empty patterns on empty and non-empty receivers, and a multibyte receiver.
- ruby/spec `core/string/{gsub,sub,match,scan}_spec.rb`, `core/regexp/{match,last_match}_spec.rb`, `core/matchdata`, `core/kernel/gsub_spec.rb`, `language/regexp`: identical to master.
- Full `cargo test` suite green; aarch64 cross-check compiles.

---

### DIFF
--- a/monoruby/src/builtins/regexp.rs
+++ b/monoruby/src/builtins/regexp.rs
@@ -1216,21 +1216,11 @@ fn rmatch(vm: &mut Executor, globals: &mut Globals, lfp: Lfp, _: BytecodePtr) ->
         // the Regexp attached through the stash above), and that object
         // *is* the result — `regexp.match(s).equal?($~)` holds in CRuby
         // — so hand it back rather than building a second MatchData
-        // with a second haystack view. The fallback only covers a
-        // frame with no svar container.
-        match vm.current_match_data() {
-            Some(md) => md,
-            None => {
-                let md = Value::new_matchdata_snap(
-                    captures,
-                    heystack,
-                    vm.resolve_haystack(heystack),
-                    regex,
-                );
-                vm.set_backref(md);
-                md
-            }
-        }
+        // with a second haystack view. A builtin always runs inside a
+        // Ruby frame, so the svar container the save went to exists.
+        let _ = captures;
+        vm.current_match_data()
+            .expect("`$~` was saved by captures_from_pos")
     } else {
         Value::nil()
     };
--- a/monoruby/src/builtins/regexp.rs
+++ b/monoruby/src/builtins/regexp.rs
@@ -1212,16 +1212,28 @@ fn rmatch(vm: &mut Executor, globals: &mut Globals, lfp: Lfp, _: BytecodePtr) ->
     };
     vm.set_match_regex(self_);
     let md = if let Some(captures) = regex.captures_from_pos(heystack, byte_pos, vm)? {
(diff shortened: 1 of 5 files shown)


sisshiki1969:
`codecov/project` reports -0.01% on `80deb17`, but the Codecov compare data shows every one of the 14 new misses is in `hashbrown/src/raw/mod.rs` (224 → 238 misses, same line count) — the vendored hash table, which this PR does not touch. The files this PR changes have identical miss counts before and after (`builtins/regexp.rs` 74 → 74, `value/rvalue/regexp.rs` 88 → 88, `value.rs` 285 → 285, `match_data.rs` 8 → 8), and `codecov/patch` passes with every changed line covered. Which hashbrown paths a run exercises varies with allocation patterns, so I'll re-run the amd64 job once the current run finishes to refresh the upload.

---

PR #1239, opened 2026-09-02
Trace each example to stderr when running ruby/spec (spec/default.mspec)

sisshiki1969:
## Changes

- `spec/default.mspec` (new): mspec config that registers a `:before` action printing each example's description to stderr right before it runs.
- `action.yml`: the `rubyspec: "true"` step also copies that file to `spec/default.mspec` in the workspace. mspec loads it automatically (it searches `.` and `spec/`), in both the `mspec` driver and the `mspec-ci` child, so rubyspec-stats' `ci.yml` needs no change.
- `doc/ruby_spec_skip_tags.md`: short note on the trace.

## Why

mspec's own `--timeout` is a watchdog thread inside the target process; it cannot fire while monoruby's main thread blocks in the kernel (e.g. `waitpid`), so a hung ruby/spec run in rubyspec-stats is only ever killed by the outer `timeout(1)`. That leaves an empty results file and no record of which example hung — the two recent core hangs (`process/waitall_spec.rb`, `process/fork_spec.rb`) had to be located by counting `--marker` dots, which only identifies the file. With the trace, the last description in the CI log is the hanging example, ready to be put in a tag file.

The config does not set `:tags_patterns`, so tag lookup keeps mspec's built-in `spec/tags/ruby/...` mapping.

## Verification (local, nightly binary)

- Normal run of `process/abort_spec.rb`, `process/fork_spec.rb`, `array/first_spec.rb`: descriptions in stderr, `stats.yml` intact (29 examples, 0 failures), fork specs still fully tagged out.
- Simulated hang (spec with `sleep 100`, `timeout -k 5 5`): exit 137, empty `stats.yml`, and `ZZ trace hangs forever` as the last stderr line.

Cost: one line per example (core ≈ 23k lines in the step log).

https://claude.ai/code/session_017gxL6QM3ZPtgV14QddDq1W

---
_Generated by [Claude Code](https://claude.ai/code/session_017gxL6QM3ZPtgV14QddDq1W)_

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -7,8 +7,9 @@
 #     - run: monoruby my_script.rb
 #
 # With the `rubyspec: 'true'` input the action also prepares a ruby/spec
-# workspace (clones spec/ruby + spec/mspec and copies this repo's spec/tags
-# into the tag locations mspec resolves), ready for
+# workspace (clones spec/ruby + spec/mspec, copies this repo's spec/tags
+# into the tag locations mspec resolves, and installs this repo's
+# spec/default.mspec, which mspec picks up by itself), ready for
 # `spec/mspec/bin/mspec ci --target monoruby spec/ruby/...`.
 #
 # Install strategy, in order:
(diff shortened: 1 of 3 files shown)


sisshiki1969:
`codecov/project` is red (92.44%, -0.02%, 17 fewer hits on an unchanged line count), but this PR touches no Rust source — only `action.yml`, `spec/default.mspec`, and a doc. The delta is run-to-run coverage variance in the existing test suite, so there is nothing to fix here. `codecov/patch`, both `smoke` jobs, and `amd64` are green; `darwin` is still running.

---

PR #1250, opened 2026-09-03
rubyspec-stats watch: fix the issue text for placeholders and untraced logs

sisshiki1969:
## Changes

Two wording fixes in `.github/scripts/rubyspec_stats_watch.rb`, found on the first real issue it opened (#1249):

- The tagging hint used `<file>` / `<description>` placeholders, which GitHub's HTML sanitizer strips, leaving `spec/tags/core/_tags.txt` and `critical(hangs):`. Now `FILE_tags.txt` / `DESCRIPTION`.
- A job log without the example trace (runs before `spec/default.mspec` from #1239 was installed, such as upstream run #1427) ends in `--marker` dots like any other log, and was described as "the process died while loading the next spec file". That wording is now used only when the log does contain trace lines; otherwise the issue says there is no trace and the dots have to be counted.

## Verification

Synthetic step logs for the three cases (dots only; trace ending in a description; trace ending in a dot) and the real pre-trace fork log; the offline issue-flow harness still passes.

https://claude.ai/code/session_017gxL6QM3ZPtgV14QddDq1W

---
_Generated by [Claude Code](https://claude.ai/code/session_017gxL6QM3ZPtgV14QddDq1W)_

### DIFF
--- a/.github/scripts/rubyspec_stats_watch.rb
+++ b/.github/scripts/rubyspec_stats_watch.rb
@@ -99,9 +99,15 @@ def step_tail(log, category)
   block = block[(endgroup + 1)..] if endgroup
   exit_line = block.find { |l| l.start_with?('##[error]Process completed') }&.sub('##[error]', '')
   output = block.reject { |l| l.empty? || l.start_with?('##[') || l.start_with?('Warning: failed to read library path') }
+  # The first two output lines are mspec's command echo and monoruby's banner;
+  # anything after that which is not a bare `--marker` dot run is the example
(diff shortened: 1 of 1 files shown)


sisshiki1969:
`codecov/project` is red (92.46%, -0.01%), but this PR only edits `.github/scripts/rubyspec_stats_watch.rb`; no Rust source or test changed, so the delta is run-to-run coverage variance of the existing suite (same as on #1239), nothing to fix here. `amd64` and `darwin` are green.

---

PR #1258, opened 2026-09-04
Compare a String subclass Hash key with eql?, not with bytes

sisshiki1969:
A Hash key that is a `String` **subclass** was compared by byte content, so a redefined `eql?` on it was never consulted:

```ruby
class Sub < String; def eql?(o) = false; end
h = {}; h[Sub.new("q")] = 1
h[Sub.new("q")]      # CRuby nil, monoruby 1
```

## Where CRuby draws the line

`rb_any_cmp` gates its String short-circuit on `RBASIC(a)->klass == rb_cString` for **both** operands; a subclass falls through to `rb_eql`, so `eql?` decides. Its `any_hash` has **no** such gate — a subclass still hashes by content.

Both halves matter, and they pull in opposite directions:

| | hashes by | compares by |
|---|---|---|
| plain `String` | content | bytes |
| `String` subclass | content | `eql?` |

So the two keys *do* land in the same bucket, and it is the redefinition that turns the lookup into a miss rather than the bucket choice. Gating only the hashing instead (my first attempt) makes `Sub` stop finding its own entry even with the default `eql?` — the four shapes in the test pin exactly this:

```
                    h[Sub.new("q")]   h["q"]
hash only               :hit           :hit     ← content still chose the bucket
eql? only               nil            :hit     ← eql? said no
neither                 :hit           :hit
both                    :hit           :hit
```

## The change

The digest stays content-based; only the comparison moves. Three places had the same unguarded short-circuit:

- **`Value::eql`'s `(STRING, STRING)` arm**, which every general probe reaches — now requires both classes to be exactly `String`, else falls through to the `eql?` dispatch already sitting below it.
- **the boxed map's vm-free fast path**, selected per probe key.
- **`is_inline_key`**, so a subclass is never *stored* inline — its equality is not answerable without the vm.

`inline_pos` grows the case that follows from that: a subclass **probe** against inline-stored plain Strings. It cannot take the byte scan, and it is not "a heap key that can never match" either — the bytes chose the bucket — so it scans the pairs with the dispatching comparison.

## Cost

None measurable. `Value::is_plain_rstring_inner` answers the type and the class from one `try_rvalue`, comparing a word the type check already loaded.

Pure-lookup micro (18-entry Hash, loop baseline subtracted), four alternating rounds:

| | master | this PR |
|---|---:|---:|
| String 12B | 36.8 | 38.1 |
| String 19B | 35.4 | 36.5 |
| String miss | 29.2 | 29.8 |
| Symbol | 24.1 | 23.6 |
| Integer | 40.2 | 34.6 |

The String rows move ~3 %, but `Symbol` and `Integer` — which this change cannot touch — move as much or more in the same runs, so the noise floor on this host is ±10-15 % and the String difference is not resolvable. Worth a confirming run on a quiet machine if the Hash path is under active tuning.

## Tests

`string_subclass_key_dispatches_eql` in `tests/hash_string_keys.rs`: all four redefinition shapes, in both representations (1 pair inline, 18 boxed), in both directions (subclass key probed with a plain String and vice versa), through `[]` / `key?` / `delete`. Verified to fail without the fix.

Full `cargo test` with `stress-spill-pool` passes (76 binaries).

Also drops a dead closure (`ivar_of`) left behind by #1254's refactor.

## How this came up

Investigating whether caching a frozen String's digest in `RStringInner` (`doc/hash_optimization.md` §4.5) is sound, since a redefined `String#hash` would invalidate a cache. It is sound — both implementations ignore a redefined `String#hash` for Hash bucketing — but checking that turned up this adjacent `eql?` divergence.

---

### DIFF
--- a/monoruby/src/codegen/jitgen/compile/method_call.rs
+++ b/monoruby/src/codegen/jitgen/compile/method_call.rs
@@ -1363,12 +1363,6 @@ impl<'a> JitContext<'a> {
                 .get_ivarid(name)
                 .filter(|ivarid| ivarid.is_inline())
         };
-        let ivar_of = |this: &Self, v: frameless::LeafValue| -> Option<Option<IvarId>> {
-            match v {
-                frameless::LeafValue::SelfIvar(name) => resolve(this, name).map(Some),
-                _ => Some(None),
-            }
-        };
         // The arithmetic and the comparison below are emitted as machine
         // instructions with no runtime check, which is only licensed while
         // the operator is still the built-in one. Take that licence for
--- a/monoruby/src/value.rs
+++ b/monoruby/src/value.rs
@@ -487,7 +487,19 @@ impl RubyEql<Executor, Globals, MonorubyErr> for Value {
                         (ObjTy::COMPLEX, ObjTy::COMPLEX) => {
                             lhs.as_complex().eql(rhs.as_complex(), vm, globals)?
                         }
-                        (ObjTy::STRING, ObjTy::STRING) => lhs.as_rstring() == rhs.as_rstring(),
+                        // Byte equality only between two *plain* Strings.
+                        // CRuby's `rb_any_cmp` gates its String
+                        // short-circuit on `RBASIC(a)->klass == rb_cString`
+                        // for both operands; a subclass falls through to
+                        // `rb_eql` below, so a redefined `eql?` on it is
+                        // observed. (`any_hash` has no such gate, so a
+                        // subclass still hashes by content — only the
+                        // comparison moves.)
+                        (ObjTy::STRING, ObjTy::STRING)
+                            if self.class() == STRING_CLASS && other.class() == STRING_CLASS =>
+                        {
+                            lhs.as_rstring() == rhs.as_rstring()
+                        }
                         (ObjTy::ARRAY, ObjTy::ARRAY) => {
                             let lhs_ary = lhs.as_array();
                             let rhs_ary = rhs.as_array();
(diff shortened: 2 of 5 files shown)


sisshiki1969:
`codecov/project` is red at **92.51 % (−0.02 %)**. Standing down on it, and I have re-run the workflow once — reasoning below so it can be checked rather than taken on trust.

**It is not the patch.** Codecov's own verdict on the diff is *"All modified and coverable lines are covered by tests"*. The delta is 23 lines that went from hit to miss **outside** the diff (Hits −13, Misses +23, Lines +10).

**The diff cannot have made those unreachable.** Outside a String *subclass* key — which nothing in the suite used before this PR added a test for it — every path routes exactly as before:

- `is_inline_key`: `is_plain_rstring_inner` and `is_rstring_inner` agree for a plain String, so inline storage is unchanged.
- `inline_pos`: the new arm is `else if k.is_rstring_inner().is_some()`, reached only when the first arm declined — i.e. only for a subclass.
- `Value::eql`'s `(STRING, STRING)` arm: the added guard holds for two plain Strings, so the same arm is taken.

**The observed noise band on this pipeline is the same size.** Misses moved by −12 on #1257 and −24 on #1254, on changes with no more reason to shift unrelated coverage than this one. `bin/test` measures under llvm-cov across the benchmarks, optcarrot and ruby-spec, so which GC and scheduler paths execute is not fixed run to run.

I could not attribute the 23 lines to anything in the diff, which is why this is a stand-down and not a fix. If the re-run lands on the same number, that is worth treating as a real signal and I will look again — the honest position is that I have shown the diff does not explain it, not that I have proven the pipeline wrong.

---

sisshiki1969:
Re-run reproduced it (−0.02 % again, Misses +21 where the first run said +23), so as promised I looked again — and this time I could attribute it. Codecov's compare API gives the per-file split:

```
total misses  +21   (4 files moved)

  +26  hashbrown/src/raw/mod.rs
   -4  monoruby/src/codegen/jitgen/compile/method_call.rs
   -2  monoruby/src/scheduler.rs
   +1  monoruby/src/builtins/kernel.rs
```

**The whole delta is `hashbrown/src/raw/mod.rs`** — the vendored table, which this PR does not touch. My own source moved the other way: `method_call.rs` −4 is the dead closure this PR deletes.

`raw/mod.rs` is the generic `RawTable` — find, insert, reserve, resize, rehash, the group scan — so which of its lines execute is a function of the table sizes and probe sequences actually reached, across the benchmark / optcarrot / ruby-spec phases of the coverage run. That is the one file in the project whose coverage is a property of the *workload*, not of the call graph.

Two further facts, for whoever reads this next:

- the two runs of the **identical commit** differ by 2 misses, so the pipeline is not deterministic at this scale;
- the head report is a merge of 2 sessions against the base's 1, and still lands lower — which is what run-to-run correlation on the same code would look like, not a coverage regression the diff caused.

I am not going to chase a vendored generic table's line coverage, and I will not pad the suite to move it. Standing down here for real: the patch is fully covered, the delta is attributed, and nothing in it is this PR's to fix.

---

PR #1262, opened 2026-09-05
GC: promote Rational, Range and Complex — and settle what is left

sisshiki1969:
Finishes the promotion allowlist started in #1256 / #1259, and answers the two questions those left open. The code change is small; most of this PR is the measurement that says what to do next — and, in one case, what *not* to do.

## 1. Rational, Range, Complex

Auditing the three turned up that they are not the same kind of case at all:

* **`RationalInner`** is `{ num: BigInt, den: BigInt }` and its `mark` is a no-op — so it is not a "stores already barriered" case, it is a reference-free kind like Time/Regexp/UnboundMethod. `mark_children` dispatches to `self.as_rational().mark(alloc)`, which is what hid it: the call looks like a traversal, the body is empty.
* **`RangeInner`** holds `start` and `end`, written in exactly one place — `Range#initialize` — which already barriers both. `initialize`'s own doc comment says the barrier is the caller's responsibility, and `builtins/range.rs` does it.
* **`ComplexInner`** holds two `Real`s (a `repr(transparent)` `Value`, so a Bignum or heap Float component is a real heap reference), but there is **no mutable accessor for it anywhere** — a Complex is built and never written again.

So no new barrier: `is_promotable` gains the three, `young_child_exists` gains arms mirroring each `mark`.

| bench | RANGE marked/minor | COMPLEX |
|---|---|---|
| erubi | 29 → 0 | 1 → 0 |
| activerecord | 74 → 2 | 1 → 0 |
| rack | 37 → 0 | 1 → 0 |
| graphql | 28 → 1 | 1 → 0 |

**The time effect is nil**, and smaller than #1259's already small one — minor mark goes erubi 984 → 1088 µs, activerecord 2281 → 2239, rack 938 → 925, graphql 969 → 973; the direction does not even agree between benchmarks. That is what removing 30–74 objects from a 6k–30k minor traversal looks like. Landing it to finish the allowlist, not for a number: like #1259 it adds no invariant anyone has to maintain.

### "Range is frozen, so it cannot be written" — worth checking, and not true

Only a *direct* `Range` is frozen, and only at the **end** of `#initialize`. `Range.allocate` returns an uninitialized, unfrozen Range, so *allocate → survive three collections → promote → initialize stores young values into an old object* is a real window; subclass instances never freeze at all.

So I checked that the existing barrier is load-bearing rather than assuming it. Builds identical but for `Range#initialize`'s `write_barrier(start)` / `(end)`, eight gc-stress runs each:

| build | result |
|---|---|
| barriers removed | **8/8 `DEAD RVALUE` abort** |
| barriers present (this PR) | 0/8 |

The repro is fussy in a way worth recording: the Strings handed to `initialize` must not stay in a local **or** in the caller's argument slots, or the frame root keeps them alive and the barrier looks unnecessary. Doing the store inside a method that then returns is what makes the window observable — my first two attempts measured nothing for exactly that reason.

## 2. Proc / Frame: stuck young, not short-lived

#1259 left this tier with an open question — is erubi's 2693 Procs/minor a promotion problem, or are they genuinely short-lived (the template allocates a block per render)? Barriering frame slots is a trade-off against the hottest path in the system, so it is worth knowing before anyone starts.

Measured by aging every survivor (promotion still gated on `is_promotable`) and taking the age distribution of what each minor marks:

| | total/minor | age 0 | 1 | 2 | 3 | 4 | 5 | **6+** |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| erubi PROC | 2697 | 18 | 18 | 18 | 18 | 0 | 18 | **2604 (96.5%)** |
| erubi FRAME | 2696 | 18 | 18 | 18 | 18 | 0 | 18 | **2603** |
| AR PROC | 2949 | 41 | 39 | 41 | 42 | 5 | 39 | **2737 (93%)** |
| AR FRAME | 2868 | 40 | 39 | 41 | 41 | 4 | 39 | **2662** |

The promotable kinds are the control: STRING / OBJECT / HASH / ARRAY all sit at ages 0–2 with **zero** at age 3+, because at 3 they promote and stop being re-marked. That the control comes out that way is what says the measurement reads what it claims to.

So the 82% of erubi's minor traversal is not garbage being looked at on its way out — it is long-lived closures rescanned forever because they cannot be promoted. The headroom is real.

What blocks it, from `Lfp::move_frame_to_heap`: promotion happens to a *running* frame (the moment a block captures it) and redirects `cfp.set_lfp(heap_lfp)`, so from then on that method's ordinary local-variable stores write the heap buffer through r14. CRuby can barrier captured environments cheaply because writes to one go through `vm_env_write`; monoruby has no such separation, so "store into a heap frame" is "every local variable store" unless one is built. §8.6 records three design options, the most promising being to promote only frames whose owner has returned — where those 2600 closures are, and where the only remaining writes are dynvar stores from nested blocks, rare enough to barrier cheaply.

## 3. "Can a promoted parent cascade OLD onto its children?"

Mechanically yes, and sound if restricted to `is_promotable()` children. Measurement says there is nothing to win, so it is written up (§8.7) rather than built:

* remember-on-promote already converges to the same state about three collections later — the parent is remembered, `mark_remembered` marks the children each minor, they age and promote at 3, `reclassify_remembered` drops the parent;
* the remembered set is cheap: 11 / 17 / 13 entries per minor on erubi / rack / graphql, 1–2% of the minor mark;
* activerecord's **158 entries, 261 µs (11%)** is the one real cost, and cascading cannot reach it — those are not parents being promoted now, they are long-since-old containers taking new young stores, i.e. the write barrier working as designed. At 1.65 µs per entry the cost is the size of what is rescanned, not the count, which points at card marking — a different design.

Cascading also bypasses the age-3 threshold, whose whole purpose is to not promote objects that merely survived once.

## Verification

`gc-debug` and `gc-stress` together — a collection at every safepoint, with #1256's clean-`ClassInfo` invariant asserted on every minor: 2618 lib tests, no assertion, no dead-object report. Ordinary suite after rebasing onto current master: 3737 passed / 1 failed (the pre-existing `float::tests::angle`, a local CRuby 4.0.6 vs 4.0.2 `Float#ceil(15)` difference).

`doc/yjit_bench_slow_investigation_2026-09.md` §8.6–8.7 carry the census, the barrier experiment, and the cascade analysis.

---

---


sisshiki1969:
`codecov/project` is red again (92.51%, −0.01% vs `1971fc1`). Same situation as #1259, and the evidence there applies directly, so I am not re-running the whole analysis — just the short version.

`codecov/patch` is green: all modified and coverable lines are covered. The −0.01% is elsewhere in the tree.

The repo has no `codecov.yml`, so codecov's default `project` status fails on *any* drop, and this repo's coverage is not stable to that precision. Measured on #1259 by running `cargo llvm-cov --release --lib` twice on **the same commit** of master, no changes in between:

```
run 1: 148044/164018
run 2: 148019/164018   Δ = -25 hits
```

with the drift landing in `hashbrown/src/raw/mod.rs` (−15), `parser/prism_backend.rs` (−10), `value/rvalue/thread.rs` (−6) — the vendored hash table's probe/resize branches, thread/scheduler paths, and parser branches downstream of them, all timing- and address-dependent. A GC change perturbs exactly those, in both directions.

−25 on an unchanged commit is larger than the −10 this PR is being failed for, so the gate is not resolving a real signal here. Nothing to fix in the diff, and lifting unrelated files with tests would be gaming the metric.

`amd64` is green; `darwin` is still running and I will follow up if it fails. Locally after rebasing onto current master: ordinary suite 3737 passed / 1 failed (the pre-existing `float::tests::angle`, a local CRuby 4.0.6 vs 4.0.2 `Float#ceil(15)` difference), and `gc-debug` + `gc-stress` together — a collection at every safepoint — 2618 lib tests with no assertion and no dead-object report.

As on #1259: a `codecov.yml` giving `project` a ~0.1% threshold would stop this recurring. Happy to send that as its own PR if you want it.

---

PR #1275, opened 2026-09-06
SlotState: derive the fpr file from the slot modes; freeze the specialized call-site save set

sisshiki1969:
Second slice of the `SlotState` slimming (after #1273). The fpr register file's reverse map is gone, and the suite caught a latent bug the map had been masking, which is fixed at its root here.

## The derivation

`FprAllocator` kept `vfpr: Vec<Vec<SlotId>>` — for each fpr, the slots bound to it — maintained by `fpr_add` / `fpr_remove` / `clear` / `swap` beside every `F` / `Sf` transition. It is a function of the slot modes, so it is now computed from them: `SlotState::fpr_slots(fpr)` and a one-pass `pool_occupancy()` (per pool register: occupied?, all-`Sf`?) serve the allocator's two phases, `is_fpr_vacant`, the call-site save set and the deopt write-back. `FprAllocator` keeps only what is *not* derivable: the number of ids issued (spill ids are never re-issued once vacant) and the pin set. `set_F` / `set_Sf` / `clear` no longer touch a file; `fpr_swap` only relabels; the bookkeeping the `alloc_fpr` aliasing regression lived in is gone.

## What the map was hiding

`outer_float_write_through`'s two type-flip tests returned wrong values with the derived file. The emitted code differed from master at exactly one place: the `each` call site in the owner loop saved `xmm2` on master and not on the branch, and the block compiled inside that call then read the owner's `a` where it meant `i` — a `_%2 = %2 == %3 [Float][Integer]` deopt storm, because the block's static frame-chain offsets were off by the missing 16-byte save.

The mechanism: `specialized_compile` freezes the caller's FP save set *before* the nested compile (`stack_offset = using_fpr_offset().offset()` lays out the callee's `extra` chain offsets over it), while the call emission after the compile took `get_using_fpr` *again*. In between, the callee's compile can widen a caller slot (`StoreDynVar` through the chain → `widen_outer_slot` → `invalidate_slot`, which set `S` **without** `fpr_remove`). With the map, the stale entry kept `xmm2` "occupied", so both sets agreed by accident (and the register was leaked for the rest of the compile). Derived from the modes, the second set was smaller than the first, and the callee's offsets no longer matched the frame the call actually built.

## The fix

The frozen set rides back on the compiled frame (`JitStackFrame::call_site_using_fpr`, surfaced as `SpecializedCompileResult::using_fpr`), and both specialized call sites (`compile_yield_specialized`, `specialized_iseq`) emit *that* set and record it for the resolve pass. `get_using_fpr` still runs for its GP flush and alias kill; a debug assertion checks the live set is a subset of the frozen one (it can only shrink — a suspended frame never gains a pool register). Saving a register the caller no longer needs is harmless; saving fewer than the callee was laid out over was the bug. Written up in doc/regalloc_separation.md §47.

## Verification

- `emit-asm` dumps of nine benchmarks (`app_fib`, `so_mandelbrot`, `so_nbody`, `app_aobench`, `binarytrees`, `quick_sort`, `bf`, `tarai`, `loop_whileloop`), this branch vs master, addresses and compile-time lines normalised: **all identical**.
- JIT lib tests (107), the float / block / loop integration tests (15 files, `outer_float_write_through` included), and the **full `cargo test` suite** pass.
- aarch64 cross build compiles.

---

### DIFF
--- a/doc/regalloc_separation.md
+++ b/doc/regalloc_separation.md
@@ -3183,3 +3183,48 @@ What was *not* folded, and why:
   (`wb_fpr`) emits stores in that order — a derived map would emit them in slot
   order, so the change is semantics-preserving but not byte-identical. Left for
   its own change with its own gate.
+
+## 47. The fpr file is derived from the slot modes (and the stale entry it was hiding)
+
+`FprAllocator` kept a reverse map `vfpr: Vec<Vec<SlotId>>` — for each fpr, the
+slots bound to it — maintained by `fpr_add` / `fpr_remove` / `clear` / `swap`
+alongside every `F` / `Sf` mode transition. The map is a function of the modes,
+so it is now computed from them: `SlotState::fpr_slots(fpr)` and a one-pass
+`pool_occupancy()` (per pool register: occupied?, all-`Sf`?) serve the
+allocator's two phases, `is_fpr_vacant`, the call-site save set and the deopt
+write-back. `FprAllocator` keeps only what is *not* derivable: the number of ids
+issued (spill ids are never re-issued once vacant — §46's `pick_vacant` note)
+and the pin set. `set_F` / `set_Sf` / `clear` no longer touch a file, and the
+bookkeeping that the `alloc_fpr` aliasing regression lived in is gone.
+
+**What the reverse map was hiding.** The suite caught the derivation:
+`outer_float_write_through`'s type-flip tests returned wrong values. The
+emitted code differed from master at exactly one place — the `each` call site
+in the owner loop saved `xmm2` on master and not on the branch — and the block
+compiled inside that call then read the owner's `a` where it meant to read `i`
+(the `_%2 = %2 == %3 [Float][Integer]` deopt storm: the block's static
+frame-chain offsets were off by the missing 16-byte save).
+
+The mechanism: `specialized_compile` freezes the caller's FP save set *before*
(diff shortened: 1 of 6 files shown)


sisshiki1969:
On the `codecov/project` −0.02% (patch coverage is 100%): the dip is in lines this PR does not touch, and I traced it per file.

- `hashbrown/src/raw/mod.rs` (−14 hits) — probe-sequence paths in the vendored table; data-dependent and unrelated to the change. This is most of the delta.
- `slot.rs` (+4 misses), `codegen.rs` (+3), `x86_64/compile/mod.rs` (+4), `jitgen.rs` (+1) — the `S → Sf(x)` bridge's "x occupied → load into a temp and swap" branch and the `FprSwap` lowering behind it. On master the suite reached them only through the stale reverse-map entries (a register still listed after `invalidate_slot` set the slot to `S`), which is exactly what this PR removes. The branch stays: a genuinely occupied target is still possible, no test exercises it on its merits.
- `scheduler.rs` / `kernel.rs` / `io.rs` (±1–2) — timing noise.

The other spot the first run flagged, `unbox_to_S`'s non-keeping `F` arm, had a real trigger and is now covered by a test that hands a block over an `F` local to `String#each_char` / `loop` (dfca14e8). Nothing further to fix here; amd64 and darwin are green on the current head.

---

PR #1277, opened 2026-09-06
SlotState: one write_back(slot, Keep) for the four write-back policies

sisshiki1969:
Third and last slice of the `SlotState` slimming (after #1273, #1275). Readability only; byte-identical.

## What changes

`SlotState` had four ways to put a slot's value into its frame slot, each with its own body and prose: `write_back_slot` (keep everything), `unbox_to_S` with `keep_claims == false` (drop views and claims, keep the type) and `== true` (the specialized-call demotion: keep claims, move a pool `F` to a spill home), `to_S_unguarded` (forget everything), plus `give_up_const` as the `C` arm of the second. What each mode became, what was written and what was forgotten was spread over four functions, and the GP-resident re-homing preamble was copied into three of them.

They are now one function, `write_back(ir, slot, Keep)`, whose match is the mode × policy table (reproduced in its doc comment), and one enum:

| `Keep`    | was                                       | meaning                                                          |
|-----------|-------------------------------------------|------------------------------------------------------------------|
| `All`     | `write_back_slot`                         | the slot gets the value; every view and claim stays              |
| `Type`    | `unbox_to_S(_, false)`, `give_up_const`   | views and claims go, the type stays (a block leaves the unit)    |
| `Nothing` | `to_S_unguarded`                          | everything goes: `S(Value)`                                      |
| `Claims`  | `unbox_to_S(_, true)`                     | claims and views stay; a pool `F` moves to a spill home          |

The rationale that lived in the old comments (why `Type` drops a `C` to `S(Value)` rather than the constant's type, why `Claims` writes nothing for a `C`, why a pool `F` migrates to a spill home) is kept in the doc comment per policy. The `*_state` analysis halves of the two old functions are folded in: the `Spill` record is still computed by the state transition and emitted through `ir.spill`, so analysis mode still emits nothing. Written up in doc/regalloc_separation.md §48.

## Verification

Arm for arm the transitions and emissions are the ones the four functions performed (the GP-resident flush under every policy, the resident drop only under `All`, `clear` exactly where the old bodies cleared), and:

- `emit-asm` dumps of nine benchmarks (`app_fib`, `so_mandelbrot`, `so_nbody`, `app_aobench`, `binarytrees`, `quick_sort`, `bf`, `tarai`, `loop_whileloop`), this branch vs master, addresses and compile-time lines normalised: **all identical**, side-exit page sizes included.
- JIT lib tests (112), the float / block / loop integration tests (15 files), and the full `cargo test` suite pass.

---

### DIFF
--- a/doc/regalloc_separation.md
+++ b/doc/regalloc_separation.md
@@ -3228,3 +3228,34 @@ over was the bug.
 Verified: `emit-asm` dumps of the nine benchmarks are identical to master
 (`app_aobench` included this time), the JIT lib tests and the float / block /
 loop integration tests pass, and the full `cargo test` suite passes.
+
+## 48. One `write_back(slot, Keep)` for the four write-back policies
+
+`SlotState` had four ways to put a slot's value into its frame slot, each a
+function with its own prose: `write_back_slot` (keep everything), `unbox_to_S`
+with `keep_claims == false` (drop views and claims, keep the type) and `== true`
+(the specialized-call demotion: keep claims, move a pool `F` to a spill home),
+`to_S_unguarded` (forget everything), plus `give_up_const` as the `C` arm of the
+second. Their differences — what each mode becomes, what is written, what is
+forgotten — were spread over four bodies and their comments, and the
+GP-resident re-homing preamble was copied into three of them.
+
+They are now one function, `write_back(ir, slot, Keep)`, whose match is the
+mode × policy table (reproduced in its doc comment), and one enum:
+
+| `Keep`    | was                          | meaning                                                   |
+|-----------|------------------------------|-----------------------------------------------------------|
+| `All`     | `write_back_slot`            | the slot gets the value; every view and claim stays       |
+| `Type`    | `unbox_to_S(_, false)`, `give_up_const` | views and claims go, the type stays (a block leaves the unit) |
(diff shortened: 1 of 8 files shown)


sisshiki1969:
On the `codecov/patch` 91.4% (6 lines of `write_back` not hit): these are arms the suite did not reach before the refactor either, re-counted as "new" lines because the four old bodies became one function.

Checked against master's per-line report for `slot.rs`:

- `(V, Keep::Nothing)` — was `to_S_unguarded_state`'s `LinkMode::V => Spill::Lit(nil)` arm: **miss on master**.
- `(V, Keep::Type | Keep::Claims)` — was `unbox_to_S`'s `LinkMode::V => ir.spill(Lit(nil))` arm: **miss on master**.
- the `unreachable!` arm — was the two old `unreachable!` arms: **miss on master** (by design).
- `(None, Keep::Type | Keep::Claims)` — on master `None` shared one hit line with `S` and `MaybeNone` (`S(_) | MaybeNone | None => {}`); the `None` case (an optional parameter the specialized caller did not pass, still unset at a call) now has its own line, so whether it was ever hit is not observable from the old report. It is a rare shape and I have not manufactured a test for it.

Project coverage is unchanged (92.55%), amd64 is green and darwin is running. Nothing to fix here.

---

PR #1281, opened 2026-09-06
SlotState: mark outer float reads at the consumption; drop the pending queue

sisshiki1969:
The last of the `SlotState` fields that #1273 (doc §46) had left in place. Byte-identical.

## What changes

The stage-A report queue (`pending_outer_float_reads`) existed because the raw-f64 consumption sites — `load_fpr_state` and the float binop helpers — were `AbstractFrame` methods that only see their own frame, so the owner frame of a dynvar-loaded value could not be marked there and the `(outer, src)` pair waited for the next `compile_instruction` boundary, where the `JitContext` holds the chain.

The consumption sites are on the chain now:

- `binop.rs`'s single `impl AbstractFrame` block is `impl AbstractState`, and `load_fpr` / `load_fpr_state` moved with it. Their bodies are unchanged — every field and frame-level method they use resolves through `AbstractState`'s `Deref`/`DerefMut` to the innermost frame. (Three `AbstractFrame::fold_cmp` references in the numeric builtins follow.)
- `AbstractState::use_as_float` does what the drain did — look up the slot's `dynvar_src`, resolve `outer` against the chain, `mark_outer_float_read` — right at the consumption. The frame keeps only the liveness half (`use_as_float_liveness`).
- The queue, its drain at the top of `compile_instruction`, its join concatenation and `take_pending_outer_float_reads` are gone.

`SlotState` is down to `slots`, `fpr_alloc`, `gp_regfile`, `local_num`.

## Why the timing change is safe

`outer` is resolved against the chain as it stands at the consumption, which is the chain the `LoadDynVar` recorded the provenance under (a nested compile pushes and pops its frames inside the same instruction; a `LoadDynVar`'s consumer is a later instruction of the same frame). The mark is a monotone hint whose readers run at merges and boundaries, after the instruction that would have drained it, so seeing it one instruction earlier changes nothing they compute. A read consumed by a block-ending instruction now rides into the join on the owner frame (ORed) instead of on a queue the merge concatenated and the successor drained — the same bit either way. Written up in doc §50.

## Verification

- `emit-asm` dumps of nine benchmarks vs master, addresses and compile-time lines normalised: eight **identical**; `app_aobench` differs only at the `%13 + %14 [<INVALID>]`/`[Float]` site that master alone emits differently across runs (the nondeterminism noted in #1273).
- JIT lib tests (113), the float / block / loop integration tests (15 files), and the full `cargo test` suite pass; aarch64 cross build compiles.

---

### DIFF
--- a/doc/regalloc_separation.md
+++ b/doc/regalloc_separation.md
@@ -3272,3 +3272,38 @@ loop, and its M1 A/B (§27.3-2c) was never run. The `L`-collection timing
 finding (§42: the multi-iteration fixpoint makes the back-edge available at
 merge time) stays in the record for whoever revisits loop-aware allocation;
 the code it justified no longer earns its field in the per-path state.
+
+## 50. `pending_outer_float_reads` removed: the mark lands at the consumption
+
+§46 kept the stage-A report queue on the grounds that the raw-f64
+consumption sites (`load_fpr_state` and the float binop helpers) were
+`AbstractFrame` methods that only see their own frame, so the owner frame of a
+dynvar-loaded value could not be marked there and the pair had to wait for the
+next `compile_instruction` boundary, where the `JitContext` holds the chain.
+
+The cheaper move is to put those consumption sites on the chain. `binop.rs`'s
+single `impl AbstractFrame` block is now `impl AbstractState`, and `load_fpr` /
+`load_fpr_state` moved with it; every field and frame-level method they use
+still resolves through `AbstractState`'s `Deref`/`DerefMut` to the innermost
+frame, so the bodies are unchanged. `AbstractState::use_as_float` then does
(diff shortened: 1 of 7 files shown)


sisshiki1969:
On the `codecov/project` −0.02% (patch coverage 100%): per file, every additional miss is in the vendored `hashbrown` (`raw/mod.rs` +14, `map.rs` +6) plus one line of `rubymap` — data-dependent probe paths that swing from run to run, as on #1275. The monoruby files this PR touches (`slot.rs`, `read_slot.rs`, `binop.rs`, `compile.rs`, the two numeric builtins) have no new misses; their hit counts only move by the lines removed or relocated. amd64 and darwin are green; nothing to fix here.

---

PR #1297, opened 2026-09-09
Builtin and stdlib gaps behind the ruby-bench failures: String#concat, Ractor, Date._parse, gzip, YAML scalars

sisshiki1969:
Built-in class and standard-library gaps behind the remaining ruby-bench failures (erubi-rails / lobsters / shipit, knucleotide-ractor, sequel, mail, rubocop, liquid-render / liquid-compile). Stacked on the priority-1 PR (triage priorities 2 and 3).

## `String#concat` recursed through `<<` (ActiveSupport::SafeBuffer)

The Ruby-level `String#concat` appended with `self << x`. `SafeBuffer` aliases `<<` to its own `concat` and calls `super` from there, so `concat → << → concat → …` overflowed the stack on every Rails boot. `concat` now appends through a hidden `String#__shl` primitive (the same Rust function as `<<`, registered under a name no subclass aliases). The non-CRuby `String#append` alias is dropped.

## `Ractor` (knucleotide-ractor)

`builtins/ractor.rb` emulates Ractor on monoruby's green threads: `Ractor.new(*args) { }`, `#value` / `#take` / `#join`, `#send` / `Ractor.receive` (inbox), `Ractor.yield` (outbox), `.current` / `.main` / `.main?` / `.count`, `make_shareable` (deep freeze, `copy:`) / `shareable?`, `RemoteError` with the original exception as `cause`, `#name` / `#inspect`. No parallelism and no isolation errors — objects are shared by reference — but a program that just wants its Ractors to compute runs.

## `Date._parse` and friends (sequel, mail)

`date_core.rb` had ISO-only `Date.parse` / `DateTime.parse` and no `Date._parse`, which `Time.parse` calls directly. `Date._parse` is now a transcription of date_parse.c's sub-parsers (weekday, time with meridian / fraction / zone, eu, us, iso, jis, vms, slash, dot, iso2 week/ordinal, year, mon, mday, digit runs, BC, 2-digit-year completion, `zone_to_diff` with the common zone table) — on a corpus of 57 inputs it produces the same hash as CRuby. `Date.parse` / `DateTime.parse` complete the fragments like `rt_complete_frags`; `DateTime` gains `offset` (Rational), `zone`, `sec_fraction`, `new_offset`, `iso8601`, a zone-aware `to_time` / `to_s` / `<=>` / `strftime`. `Date#wday` was off by one (JD 0 is a Monday), which the ISO week-date completion exposed.

## gzip framing in `Zlib` (rubocop → unicode-display_width)

`Zlib.gzip` / `Zlib.gunzip`, `GzipReader` (header fields, CRC-32 / ISIZE trailer checks, `read` / `gets` / `each_line` / `readlines` / `getc` / `getbyte` / `ungetc` / `rewind` / `unused`, `.open` / `.wrap` / `.zcat`) and `GzipWriter` (`write` / `<<` / `puts` / `print` / `putc`, `mtime=` / `orig_name=` / `comment=`, `.open` / `.wrap`). `Inflate.new(window_bits)` / `Deflate.new(level, window_bits)` honour zlib's framing selection (negative = raw, `+16` = gzip, `+32` = auto-detect). The Inflate block loop was split out (`__blocks_at`) so raw and gzip streams share the decoder.

## YAML multi-line flow scalars (liquid's vision database)

`psych.rb` kept the line breaks of a quoted or plain scalar that continues on following lines (`content:\n  "<p>…</p>\n  <p>…</p>"`), and did not unescape `''` in single-quoted scalars. Continuation lines now fold into single spaces as in YAML 1.1, whether the value starts on the key's line or the next one.

## Reported Unicode version (rubocop → unicode-emoji)

The vendored rbconfig snapshot reported the host CRuby's `UNICODE_VERSION` (17.0.0), but the bundled Onigmo ships the Unicode 12.1 property tables — it knows `\p{Emoji_Presentation}` but not the later aliases (`\p{EPres}`), so unicode-emoji picked its "native" regexps and died at load with an invalid property name. `startup.rb` now reports the bundled tables' version (12.1.0 / emoji 12.1) so such gems take their portable tables.

## Tests

`string_concat_does_not_dispatch_through_shl`, `ractor.rs` (2), `date_parse.rs` (2: the corpus, and `Time.parse` / `DateTime.parse` round trips), `zlib_gzip.rs`, `yaml_scalars.rs`. All compared against CRuby 4.0.2. Full `cargo test` green on the stack.

ruby-bench after this PR: knucleotide-ractor, sequel and mail pass; erubi-rails / lobsters / shipit get past the SafeBuffer overflow and stop at nokogiri, rubocop gets past GzipReader / Marshal / unicode-emoji and stops at the prism gem — both C extensions, out of reach.

---

### DIFF
--- a/monoruby/builtins/ractor.rb
+++ b/monoruby/builtins/ractor.rb
@@ -1,0 +1,351 @@
+# Ractor — a sequential-consistency emulation on top of monoruby's green
+# threads.
+#
+# monoruby has no parallel execution and no object-isolation machinery;
+# every Ractor here is a Thread with an inbox (`#send` / `Ractor.receive`)
+# and an outbox (`Ractor.yield` / `#take`), and the block's return value is
+# the thread's value (`#value` / `#take` after completion). Objects are
+# passed by reference, never copied or moved; `Ractor.make_shareable`
+# deep-freezes as CRuby does, so code that relies on it still gets frozen
+# data, but the isolation errors CRuby raises for unshareable captures are
+# not reproduced.
+#
+# What is covered is the API surface a program written for real Ractors
+# touches when it just wants its work done: `Ractor.new(*args) { |*args| }`,
+# `#value`, `#take`, `#join`, `#send` / `#<<`, `Ractor.receive` /
+# `Ractor.recv`, `Ractor.yield`, `Ractor.current` / `.main` / `.main?` /
+# `.count`, `Ractor.make_shareable` / `.shareable?`, `#name`, `#inspect`.
+class Ractor
+  class Error < StandardError; end
+  class IsolationError < ArgumentError; end
+  class MovedError < Error; end
+  class ClosedError < StopIteration; end
+  class UnsafeError < Error; end
+
+  # Raised by `#value` / `#take` when the block raised; the original
+  # exception is `#cause`, as in CRuby.
+  class RemoteError < Error
+    attr_reader :ractor
+
+    def initialize(ractor, message = "thrown by remote Ractor.")
+      super(message)
+      @ractor = ractor
+    end
+  end
+
+  class MovedObject < BasicObject
+    def method_missing(*)
+      ::Kernel.raise ::Ractor::MovedError, "can not send any methods to a moved object"
+    end
+
+    def respond_to_missing?(*)
+      false
+    end
+  end
+
+  @count = 0
+  @main = nil
+  @sequence = 0
+
+  class << self
+    # Every Ractor other than the main one runs its block on its own
+    # (green) thread; the thread's value is the Ractor's value.
+    def new(*args, name: nil, &block)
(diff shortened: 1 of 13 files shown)


sisshiki1969:
Rebased onto master after #1296 was squash-merged (head d7ee16ab) and retargeted the PR to master. Retargeting does not fire the `pull_request` workflow (only opened / synchronize / reopened do), so this head has no CI run yet; the stack's tests (`string_concat`, `ractor`, `date_parse`, `zlib_gzip`, `yaml_scalars`, `gem_prelude`, `require`, `predefined_gvar`, `language_fixes`) pass locally on the rebased tree, and the full suite was green before the rebase. CI will run on the next push to the branch.

---

PR #1298, opened 2026-09-09
Keep installed-gem lib dirs out of $LOAD_PATH; resolve them as a require fallback

sisshiki1969:
monoruby seeded `$LOAD_PATH` with every installed gem's `lib/` (the runtime probe's list), ahead of anything `Bundler.setup` later prepends. CRuby never lists gem directories there: rubygems activates a gem on demand from its `require` hook and only then splices the `lib/` in. Two consequences showed up in ruby-bench's liquid-render / liquid-compile (triage priority 3; stacked on the builtins PR):

- `$LOAD_PATH.detect { |p| File.exist?(File.join(p, "liquid.rb")) }` found the host's liquid-5.13.0 gem instead of the bundle's git checkout (`performance/theme_runner` does not exist there → LoadError).
- `Bundler.setup` could not remove the stray entries (`clean_load_path` only knows paths of activated specs), so any gem installed on the host shadowed the bundled version of the same gem.

Now the probe's list is partitioned at startup: entries under a `GEM_PATH` root's `gems/` or `bundler/gems/` go to `Globals::gem_lib_dirs`, everything else stays in `$LOAD_PATH`. `require` consults the gem dirs only after `$LOAD_PATH` misses, and a hit is spliced into `$LOAD_PATH` (like an activation) so the gem's other files resolve through the ordinary walk from then on. `$LOAD_PATH` at boot now has CRuby's shape (stdlib + site/vendor dirs); the gem-prelude / require / rubygems-activation / predefined-gvar tests are unchanged and pass.

ruby-bench after this PR: liquid-render and liquid-compile pass (172 ms / 49 ms per iteration here).

---

### DIFF
--- a/monoruby/src/globals.rs
+++ b/monoruby/src/globals.rs
@@ -348,6 +348,16 @@ pub struct Globals {
     encoding_objects: HashMap<crate::value::rvalue::Encoding, Value>,
     /// library directries.
     load_path: Value,
+    /// `lib/` directories of the host's installed (non-default) gems,
+    /// consulted by `require` only after `$LOAD_PATH` misses. CRuby does
+    /// not list them in `$LOAD_PATH` either: rubygems activates a gem on
+    /// demand and only then splices its `lib/` in. Keeping them out of
+    /// the visible array lets `Bundler.setup` (which prepends the
+    /// bundle's paths) win over every other installed version of a gem,
+    /// and keeps `$LOAD_PATH` scans (`$LOAD_PATH.detect { … }`) from
+    /// landing on an unrelated gem. A hit here is appended to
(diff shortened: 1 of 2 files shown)


sisshiki1969:
The `codecov/project` check reports −0.01% (92.62% → 92.61%). Per the Codecov comparison, the 23 lines this PR adds (`globals.rs`, `require.rs`) are all covered; the 14 new misses are entirely in `hashbrown/src/raw/mod.rs` (the vendored hash table, untouched here), i.e. run-to-run variance in which raw-table paths the suite exercises, not a coverage loss from this change. `codecov/patch`, `amd64` (and `darwin`, still running at the time of writing) are otherwise the signals to go by; nothing to fix on this side.

---

PR #1325, opened 2026-09-12
JIT: name the kwrest table by address, not PC-relatively

sisshiki1969:
`RestKw` hands `correct_rest_kw` a table of (name, slot-id) pairs, and on aarch64 that address went in with `adr`. It reaches ±1 MiB, and constants are emitted at finalize behind the whole compilation unit, so a `**kwrest` call site early in a large unit could not name its own table.

That is the overflow #1324 fixed for `opt_case`'s jump tables, in the one const-area reference the branch relaxation does not cover: `far_branch_mode` never gated this path, so a large frame aborts with `ADR displacement out of range` whether or not it is in far mode.

```ruby
def sink(x, **opts) = x + opts.size
def big(v)
  a = sink(v, k: 1, j: 2)   # its table ends up ~1.8 MiB away
  # ... a megabyte of body ...
end
```

## The fix

The one the unit's class-version word already uses: resolve the label before any of the unit's code is emitted, then bake an absolute address, which has no range at all.

Building the tables moves out of both backends into a single walk of the AsmIr tree, beside the other pre-codegen rewrites. The two backends now only reference a label. x86 reads it rip-relative exactly as before; aarch64 moves it as an immediate, two instructions more than the `adr` it replaces, on a path that then calls a runtime function.

## Why the test is shaped the way it is

It lives in `tests/far_branch_relaxation.rs` with the rest of the family, but differs from the three already there in two ways the bug requires.

- It goes through the **method** JIT. The loop JIT's caller catches a panic from the emitter and leaves the loop interpreted, so a loop-shaped test passes with or without the fix. The method JIT's caller does not catch it.
- It is sized to clear the ±1 MiB reach outright, not merely to cross the relaxation threshold. Measured at 120,039 AsmInsts in one unit, about 1.8 MiB of aarch64 code at the ratio #1324's case measured, with the call site at the top.

## Verification

Full suite on x86-64: 108 test binaries, 3,522 tests, no failures. The aarch64 compile and behaviour are for the darwin job, which is what exercises the changed instruction.

https://claude.ai/code/session_013q5CpZ6eqZUJtPhSJHfjus

---
_Generated by [Claude Code](https://claude.ai/code/session_013q5CpZ6eqZUJtPhSJHfjus)_

### DIFF
--- a/monoruby/src/codegen/arch/aarch64/compile/method_call.rs
+++ b/monoruby/src/codegen/arch/aarch64/compile/method_call.rs
@@ -1127,22 +1127,21 @@ impl Codegen {
         );
     }
 
-    /// `RestKw`: build a const-data table of (name: i32, slot-id: i32) pairs
-    /// terminated by (0, 0), then call `correct_rest_kw(&table, lfp)` which
-    /// reads the listed slots and returns the `**kwrest` Hash in x0. Mirrors
-    /// the x86 `RestKw` arm; the const-table emission is arch-neutral and the
-    /// table address is taken with PC-relative `adr` (as in OptCase).
-    pub(in crate::codegen::jitgen) fn emit_rest_kw(&mut self, rest_kw: Vec<(SlotId, IdentId)>) {
-        let data = self.jit.const_align8();
-        for (i, name) in rest_kw.into_iter() {
-            self.jit.const_i32(name.get() as i32);
-            self.jit.const_i32(i.0 as i32);
-        }
-        self.jit.const_i32(0);
-        self.jit.const_i32(0);
+    /// `RestKw`: call `correct_rest_kw(&table, lfp)`, which reads the slots
+    /// the table lists and returns the `**kwrest` Hash in x0. The table of
+    /// (name: i32, slot-id: i32) pairs, terminated by (0, 0), was laid down
+    /// in the constant area by `Codegen::resolve_rest_kw_tables` before this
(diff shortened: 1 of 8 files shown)


sisshiki1969:
`codecov/project` is red at 92.87% against a 92.88% base, and the drop is not in this diff.

`codecov/patch` passes at 100%: 39 hits, 0 misses. Pulling the file-level comparison from Codecov's API, every file whose miss count moved is in the vendored hashbrown crate, which this PR does not touch and whose line counts are unchanged:

| Δ misses | Δ lines | file |
| --- | --- | --- |
| +14 | 0 | `hashbrown/src/raw/mod.rs` |
| +6 | 0 | `hashbrown/src/map.rs` |
| 0 | +28 | `monoruby/src/codegen/jitgen.rs` |
| 0 | +3 | `monoruby/src/codegen/jitgen/asmir.rs` |
| 0 | +2 | `monoruby/src/codegen/jitgen/asmir/compile_shared.rs` |
| 0 | −7 | `monoruby/src/codegen/arch/x86_64/compile/mod.rs` |

Every file this PR changes has an unchanged miss count. The 20 lines are hashbrown's grow and rehash paths, which run or not depending on insertion order and hasher seeding, so they differ between runs rather than tracking a change in the diff. 20 lines out of 172,469 is the 0.02% the check is reporting.

There is nothing in the diff to fix for it. I will spend one re-run on the coverage job once `darwin` finishes, and report what comes back.

---

PR #1328, opened 2026-09-13
JIT: emit `Foo.new` without the `Class#new` trampoline for every class

sisshiki1969:
## The problem

`Class#new` is the Ruby trampoline in `builtins/class.rb`: allocate, then forward `(...)` to `initialize`. Binding that `...` costs one empty `Array` per construction unless the JIT elides it (D1), and **D1 only fires when the trampoline is inlined into the caller's compilation unit** — which depends on the specialization-depth budget, spent from the outside in.

So the same allocation loop ran fast or slow depending on how deeply its caller happened to be nested. On ruby-bench's `string_malloc_pressure` that was 8 ms/iteration with `Class#new` inlined at the hot site and 29 ms without, the difference being exactly one `create_array` call in the innermost block (and 1.10 vs 0.77 GCs per iteration).

`inline_class_new` already emitted the construction as the caller's own instructions, with no trampoline frame and no rest `Array` — but only for a class with the stock object allocator **and** an `initialize` it could fold (a trivial body) or expand (a plain ivar-store constructor). That excluded String, Array and Hash, which is exactly what the benchmark constructs.

## The change

Widen `inline_class_new` on both axes.

**The allocator need not be the stock one.** `emit_class_allocate` already falls back to calling the class's own `alloc_func` when handed no inline payload, so pass it the same `InlineAlloc` choice `gen_class_allocate_inline` makes for a bare `Foo.allocate` — `Object` / `Struct` in line, everything else the runtime call. No new codegen; the trampoline frame goes away either way.

**A third plan for an `initialize` that is neither foldable nor expandable:** an ordinary call to it with the object just allocated as the receiver.

```rust
enum InitPlan {
    Fold,                             // trivial body: emit nothing
    Stores(frameless::IvarStoreBody), // plain constructor: expand into the caller
    Call,                             // anything else: call it on the new object
}
```

`AbstractState::send` grows a receiver override for this. The two halves belong together: the call site names the *class* as its receiver, and its `dst` holds the object rather than this call's result, so the override also discards the result.

```rust
recv_override: Option<SlotId>,
...
let dst = if recv_override.is_some() { None } else { store[callid].dst };
```

Only `set_arguments`' "fill self" actually writes the receiver (`SetupMethodFrame` reads only block info from the callid, and `AsmInst::Call` takes no receiver register), so that is the single place the override lands. The generic argument-binding arms re-read the receiver from the call site at runtime and cannot honour it, so the new leg gates itself on the same `is_simple_call` those arms are the negation of — which also lets it cover a native `initialize` (`String`, `Hash`) and one declaring optional keywords the site does not pass.

The call leg is restricted to a Ruby or native `initialize`: an attr / Struct accessor and a `define_method` proc-method are entered differently by `compile_method_call` (the latter on the *inner* block's FuncId with its definition-time outer LFP), so an `alias_method :initialize, :x=` keeps the trampoline.

### Soundness

- **GC rooting** — `def_reg2acc_class` stores the object to `dst`'s stack home before the call, so it is reachable from the caller's frame during a collection inside `initialize`.
- **Frame capture** — `send`'s trailing `immediate_evict` emits the `guard_capture` deopt when `no_capture_guard` is unset, so control never resumes here with rbp-relative reads pointing at an abandoned frame.
- **Abstract state** — `send`'s `clear_above_next_sp` may drop `dst`'s link; the home still holds the object, so it is re-established with `def_S_guarded` rather than reloaded.
- **Deopt** — between allocate and call, a deopt re-runs the whole `Foo.new` in the VM (the first object is just garbage), the same reasoning the existing `Stores` leg's hoisted frozen guard relies on. After the call, `immediate_evict` / `chain_exit` resume at `next_pc`.

Everything is in the arch-neutral front end; both backends already implement `emit_class_allocate` with a `None` payload, so there is no aarch64-side change.

## One behaviour change worth a look

On the call leg, `initialize` runs with no `Class#new` frame above it, so a backtrace taken inside it is one level shorter:

```
before                                   after
bt.rb:5:in 'BT#initialize'               bt.rb:5:in 'BT#initialize'
<internal:class>:19:in 'Class#new'       bt.rb:8:in 'Object#make'
bt.rb:8:in 'Object#make'                 bt.rb:11:in 'Object#top'
```

CRuby does the same as of 3.5's `opt_new` ([Feature #21254](https://bugs.ruby-lang.org/issues/21254)), and monoruby's `nearest_caller_location` already treats `builtins/` frames like this one as internal. But there it is a bytecode change that holds in the interpreter too, while here a cold site still pushes the frame — so the same program can print different backtraces depending on JIT state. Flagging it rather than working around it; happy to gate the leg if you would rather keep the frame.

## Measurements

Release, x86-64, median of three. Run-to-run spread on the short rows is 1–3 ms, so those are reported as unchanged.

| | before | after |
|---|---:|---:|
| the repro loop, site past the depth limit | 23.4 ms/iter | **6.2 ms/iter** |
| the same loop at depth 0 | 23.8 ms/iter | **6.3 ms/iter** |
| `String.new` ×2M | 133 ms | 121 ms |
| `String.new` nested 3 deep ×2M | 133 ms | 123 ms |
| `Hash.new` ×2M | 154 ms | 143 ms |
| `Array.new(4)` ×500k | 39 ms | 37 ms |
| `String.new("ab")` ×2M | 229 ms | 225 ms |
| non-expandable `initialize` ×2M | 300 ms | 283 ms |
| ditto, second shape ×2M | 302 ms | 287 ms |
| `Struct` instance ×2M | 271 ms | 268 ms |
| ivar-store `initialize` ×2M | 74 ms | 75 ms |
| `Object.new` ×2M | 76 ms | 76 ms |

The point is not only the 3.8x but that **the two depths now agree** — the construction costs the same wherever it sits.

The last three rows are legs this PR does not change (the `Struct` allocation was already inline via `Class#allocate`; the fold / expand legs already skipped the trampoline). They are listed to show the widened gate costs them nothing.

The micro rows improve by less than the repro loop because in a shallow `N.times { … }` the trampoline was already inlined and D1 already elided the `Array`; what is left there is one frame, which is the 5–9%.

## Validation

- `cargo test -p monoruby --no-fail-fast` — **3529 tests / 109 binaries, 0 failed**, against a live `ruby 4.0.2` (the version `vendor/ruby-stdlib/.ruby-version` pins).
- `--features gc-stress,stress-spill-pool` over `class_new_inline`, `array_initialize`, `kwargs_forward`, `forwarded_block_yield`, `forwarding_recompile`, `redefine` — **55 tests, 0 failed**. A collection at every safepoint is what proves the new object stays rooted across the `initialize` call.
- New `monoruby/tests/class_new_inline.rs` (7 tests): the native-`initialize` leg, the iseq call leg, subclasses / Struct, a site past the depth limit, the error paths, class-version invalidation, and the `define_method` decline. Re-verified with `MONORUBY_TEST_ORACLE=ruby` against 4.0.2; the oracle entries are byte-identical. The second commit records the `gc-stress` loop-count variants, as the committed file already does for `tests/method_call.rs`.
- Not measured here: ruby-bench and optcarrot are not present in this container, so the 8 ms / 29 ms figures above are from the investigation that produced this change, not re-run. The repro loop is the same structure and was measured directly.

## Docs

`builtins/class.rb`'s comment and §3.7 of `doc/arg_forwarding_jit.md` now say that the `(...)` forward is the fallback for when `inline_class_new` declines, rather than the main path for `X.new`.

---


sisshiki1969:
## Correction: this PR does not improve `string_malloc_pressure`

The description opens by citing ruby-bench's `string_malloc_pressure` (8 ms/iteration vs 29 ms). **Those numbers were carried over from the investigation that produced this change — they are not a before/after of this PR, and I did not verify them against it.** ruby-bench was not available in the container I measured in, so I used a synthesized loop of the same shape instead. Now that I have the real benchmark, it does not move:

| `string_malloc_pressure`, 700 iters | with this PR | with `inline_class_new` disabled entirely |
|---|---:|---:|
| round 1 | 9 ms | 9 ms |
| round 2 | 9 ms | 9 ms |

The reason is in the benchmark itself:

```ruby
live << Array.new(200) { String.new(capacity: 64 * 1024) }
```

`String.new(capacity:)` is a keyword call site, and `inline_class_new` declines those at its first gate (`callsite.kw_may_exists()`). So the construction this benchmark performs never reaches any of the three legs. Removing the keyword from the benchmark does not change the picture much either (7–8 ms with the PR, 8–9 ms without).

The 23 → 6 ms figure in the description is real, but it is for the synthesized loop, whose nesting differs from the benchmark it was meant to stand in for. The other rows (`String.new`, `Hash.new`, `Array.new(4)`, aobench) were measured directly and stand.

### What does move it

The cost this PR set out to remove — the rest `Array` from `Class#new`'s `(...)` forward — is better addressed by exempting forwarding callees from the specialization depth budget, leaving the Ruby `Class#new` definition to do the work:

| | this PR | + forwarding depth exemption | no inliner, exemption only |
|---|---:|---:|---:|
| `string_malloc_pressure` | 9 ms | **5 ms** | **5 ms** |
| aobench | 5.34 s | 5.38 s | 5.86 s |
| optcarrot `--opt` | no change | no change | no change |

That change is not ready to propose: it has a reproducible non-monotonic cliff (at an exemption cap of 4–5 the same benchmark runs at 41–43 ms, far worse than not exempting at all), and the mechanism is not yet understood. I am investigating it before suggesting anything.

### What stands

The soundness work in this PR is unaffected — the widened gate, the receiver override, and the test coverage all hold, and the follow-up fixing `inline_class_new`'s missing salvage record (a bug that predates this PR but whose reach it widened) is separate.

Apologies for the unverified headline number.

---

Issue #1346, opened 2026-09-14
`super` searches with the first-registered name when one proc is define_method'd under several names (doc/super_resolution.md §6)


sisshiki1969:
Filed from `doc/super_resolution.md` §6 ("Known limits"). The doc frames it as an invoker-boundary limitation (`send` / `Method#call` / Fiber), but it **reproduces on a plain method call too**, so that is how it is written up here.

## Symptom

When one proc is registered under several names with `define_method`, `super` in its body searches for the super method under the **name it was registered with first**, not the name it was called as.

```ruby
class Base
  def a = "Base#a"
  def b = "Base#b"
end
class Sub < Base
  body = ->() { "Sub(" + super() + ")" }
  define_method(:a, body)   # :a registered first
  define_method(:b, body)
end

Sub.new.a               # CRuby: "Sub(Base#a)"   monoruby: "Sub(Base#a)"  ✅
Sub.new.b               # CRuby: "Sub(Base#b)"   monoruby: "Sub(Base#a)"  ❌
Sub.new.send(:b)        # CRuby: "Sub(Base#b)"   monoruby: "Sub(Base#a)"  ❌
Sub.new.method(:b).call # CRuby: "Sub(Base#b)"   monoruby: "Sub(Base#a)"  ❌
```

Swap the registration order (`define_method(:b, ...)` first) and both `a` and `b` answer `Sub(Base#b)`. Whichever name is baked in first is the one every name then uses.

## What is and isn't affected

| shape | CRuby | monoruby |
|---|---|---|
| one proc under two `define_method` names, called by the later name | searches under the called name | ❌ searches under the first name |
| a separate block per name | `a`→`Base#a`, `b`→`Base#b` | ✅ matches |
| a normal method plus `alias_method` | both resolve to the original name (`Base#a`) | ✅ matches |

`alias_method` agrees because CRuby also searches under the original name there. The divergence is only for a proc reused across `define_method` names.

## Reproduction

```sh
cat > /tmp/sup.rb <<'RUBY'
class Base
  def a = "Base#a"
  def b = "Base#b"
end
class Sub < Base
  body = ->() { "Sub(" + super() + ")" }
  define_method(:a, body)
  define_method(:b, body)
end
puts Sub.new.b
RUBY
./target/release/monoruby /tmp/sup.rb   # Sub(Base#a)
ruby /tmp/sup.rb                        # Sub(Base#b)
```

[...]

Issue #1345, opened 2026-09-14
green threads: a kernel-blocking FFI call (sqlite3_step, …) or fcntl(F_SETLKW) stalls every green thread (doc/threads.md §10-4)


sisshiki1969:
Filed from item 4 of `doc/threads.md` §10 ("Known limitations and future work"). The doc carries the diagnosis already; there was no issue tracking it.

## Current state

Native offload (`doc/threads.md` §9) covers only **flock, FIFO open, and FFI calls marked blocking**, which leaves two gaps:

1. Other kernel-blocking operations — `fcntl(F_SETLKW)` and friends — are not covered (the doc notes there is room to add `NativeOp` kinds).
2. `submit` spawns a thread per operation rather than using a pool, so the round trip costs **60–110 µs**; anything called frequently cannot be offloaded at all.

The consequence the doc gives as its example:

> `sqlite3_step` can block for a long time waiting on busy_timeout, but it is called per row, so it is not marked blocking — **and that wait still stops the other green threads.**

In other words, one green thread waiting in the kernel stops the scheduler, i.e. every green thread in that process. For an application touching sqlite3 from several threads, the choice today is to pay 60–110 µs per row or to stall everything on a lock wait.

## What a fix looks like (per the doc)

- Turn `submit` into a **real worker pool** so the round trip is cheap enough for per-row callees — the doc's own words are that true pooling "is the prerequisite for lifting this constraint"
- On top of that, add `NativeOp` kinds to cover `fcntl(F_SETLKW)` and the rest

## On the other items in §10

Checked while filing this:

- Item 2, "a `Thread` subclass does not run a Ruby `initialize` override", is **no longer true** — `class MyThread < Thread; def initialize(*a); $ran = true; super; end; end` sets `$ran` and matches CRuby. The doc is stale there.
- Item 3, "`Thread#priority` only stores the value", still holds in the sense that it does not affect scheduling, but the observable API matches CRuby (the value set is the value returned). Its impact is not as concrete as item 4, so it is not included here.

Issue #1344, opened 2026-09-14
No TracePoint, and `set_trace_func` is a no-op stub — it needs JIT eviction on enable (doc/jit_invariants.md I7)


sisshiki1969:
Filed from `doc/jit_invariants.md` §3.9 (I7). The doc already frames it as "not merely unimplemented, but an assumption the current optimizations lean on"; there was no issue tracking it.

## Current state

`Kernel#set_trace_func` exists with CRuby's shape (private instance method and module function), but it only accepts and remembers the handler — `monoruby/builtins/kernel.rb`: *"monoruby has no line-event tracing hooks; accept and remember the handler so the method exists with CRuby's shape"*. There is no `TracePoint` class.

```ruby
ev = []
set_trace_func(proc { |event, *| ev << event })
def m; 1; end
m
set_trace_func(nil)
p ev.uniq
p defined?(TracePoint)

#   CRuby    => ["c-return", "line", "c-call", "call", "return"]
#               "constant"
#   monoruby => []
#               nil
```

(Checked on x86-64 Linux, release, `monoruby 0.3.0` / `4c28419`.)

## Why it cannot just be added

As I7 sets out, the JIT erases the very fact that a call happened:

- **trivial method fold** — a callee with `ISeqHint::ConstReturn` / `SelfReturn` loses its call entirely
- **frameless expansion** — `def initialize(a,b) = (@a=a; @b=b)` is expanded into the caller's instructions with no frame
- **specialization** — the callee is pulled into the compilation unit (the frame is still built, so backtraces survive, but the calling convention is gone)

Adding event hooks on top of that would drop `:call` / `:return` / `:line` events — the hooks would not merely change behaviour, they would lie.

## The approach the doc points at

Make "are hooks enabled" a new global invariant and evict all JIT code when they are turned on (the same shape as I3). While hooks are off the cost stays zero, which also makes it isomorphic to BOP redefinition (`doc/bop_redefinition.md`).

## Expected

- The snippet above produces CRuby's event sequence
- `TracePoint` exists and delivers at least `:call` / `:return` / `:line`
- Enabling hooks evicts JIT code, and while they are disabled the emitted code is unchanged (ideally confirmed by `--features emit-asm` output being byte-identical with hooks off)

sisshiki1969:
Scoping notes from reading the code, left here rather than starting the work — this is a staged project, not one change.

[...]

Issue #1343, opened 2026-09-14
`public_send` raises NoMethodError for a private method instead of consulting `method_missing`


sisshiki1969:
## Symptom

When `public_send` lands on a private method, CRuby hands it to `method_missing`; monoruby raises `NoMethodError` on the spot.

```ruby
class A
  private def priv = "priv"
  def method_missing(name, *a) = "mm:#{name}"
  def respond_to_missing?(n, p = false) = true
end

A.new.public_send(:priv)
#   CRuby    => "mm:priv"
#   monoruby => NoMethodError: private method 'priv' called for an instance of A
```

## What is and isn't affected

The neighbouring cases all agree, so the only divergence is "private ⇒ `method_missing`".

| expression | CRuby | monoruby |
|---|---|---|
| `A.new.public_send(:priv)` (private, `method_missing` defined) | `"mm:priv"` | ❌ `NoMethodError` |
| `B.new.public_send(:priv)` (private, no `method_missing`) | `NoMethodError` | ✅ `NoMethodError` |
| `A.new.public_send(:nope)` (undefined method) | `"mm:nope"` | ✅ `"mm:nope"` |
| `A.new.send(:priv)` | `"priv"` | ✅ `"priv"` |

So the `method_missing` dispatch works for an *undefined* method; it is only the visibility rejection that raises without going through it. In CRuby, a visibility violation found by `rb_method_call_status` still routes to `method_missing` — `NoMethodError` is what its default implementation raises.

## Reproduction

```sh
cat > /tmp/ps.rb <<'RUBY'
class A
  private def priv = "priv"
  def method_missing(name, *a) = "mm:#{name}"
  def respond_to_missing?(n, p = false) = true
end
p(begin; A.new.public_send(:priv); rescue => e; e.class; end)
RUBY
./target/release/monoruby /tmp/ps.rb   # NoMethodError
ruby /tmp/ps.rb                        # "mm:priv"
```

Environment: x86-64 Linux, release, `monoruby 0.3.0` (`4c28419`). Same on the baseline commit (`adca54e`), and the same with `--no-jit`.

Where to look: `public_send` in `monoruby/src/builtins/kernel.rs`. Instead of building a `NoMethodError` when visibility rejects the call, it should join the `method_missing` dispatch that the undefined-method case already takes.

## Context

Found while auditing the JIT's inliners; unrelated to the `send` optimization there (`public_send` deliberately gets no direct-call rewrite — its visibility check is the reason the method exists).

PR #1347, opened 2026-09-14
Two inliners off the hot page, and a `send(:literal)` that resolves at compile time

sisshiki1969:
## What the audit said

I measured every method the JIT inlines as machine code — 96 registrations — by the number of instructions each one emits at a single call site (`emit-asm`, attributing machine code to its bytecode marker) and by the time the same loop takes with the inliner on and off.

The first thing that came out of it reframes the rest: **a call site that does *not* inline is not free.** A plain builtin call emits 31–39 instructions of guard + frame + `call` + return (measured: `Integer#[]` with a variable index 35, `String#==` on two Strings 39, a Ruby-defined method's call site 31). So an inliner under ~35 instructions is *smaller* than the call it replaces, and 90 of the 96 are. Dropping them would grow the hot page, not shrink it.

These exceeded that budget:

| site | instructions | vs. the alternative |
| --- | ---: | --- |
| `Hash#[]`, String key | 204 | ~1.2–1.4x over the direct helper call |
| `Hash#[]`, Symbol key | 153 | 1.6x over the direct helper call |
| `Kernel#send(:literal)` | 97 | 11x over a full method call — but **4.7x slower than the same call written out** |
| `String#<<`, String arg | 86 | 3.2x over the direct helper call |
| `String#setbyte` | 84 | 3.5x over a full method call |
| `String#<<`, Fixnum arg | 48 | 6.4x over the direct helper call |

`String#<<` earns its size (dropping the fast paths for the frame-skipping helper call alone costs 3–6x), so it stays. The other three are what this PR is.

## 1. The hot page was carrying cold code

Two of the three were not "too much fast path" — they were a rarely-taken block sitting in the middle of the hot one.

- **`gen_hash_probe`** emitted both regimes inline: the linear walk (every map still in its `ar_table` form, `AR_MAX` = 8) *and* the hashbrown control-group walk, ~90 instructions of SWAR scan unrolled over two words. The indexed regime now goes behind `select_page(1)`; the linear one keeps the hot page.
- **`emit_string_setbyte`** put its shared-receiver detach in line — `save_registers` / `restore_registers` alone are ~50 instructions against a ~34-instruction fast path. The comment already called the block "out of line"; now it is.

| site | before | after |
| --- | ---: | ---: |
| `Hash#[]`, Symbol key | 153 | **53** |
| `Hash#[]`, String key | 204 | **70** |
| `String#setbyte` | 84 | **35** |

No measurable speed cost, including for the paths that moved (min of 15 interleaved runs, 5M lookups): an 8-entry hash 64.1 → 58.7 ms, a 256-entry (indexed) hash 36.3 → 36.5 ms, `String#setbyte` 23.1 → 23.3 ms, and a loop that setbytes a freshly duped — i.e. shared — receiver every iteration 74.8 ms, against 96.2 ms for the same loop with the inliner declined. Total JIT hot-page bytes: −38 % on a hash-heavy script, −17 % on a string-heavy one, byte-identical on scripts that touch neither (`plb2/sudoku`, `plb2/bedcov`, `app_aobench`).

x86-64 only: aarch64's `b`/`b.cond` cannot reach monoasm's second page, which `arch/aarch64/compile/index.rs` already documents.

## 2. `send(:foo)` did not have to be a lookup at all

`o.send(:foo, ...)` with a literal Symbol ran the full runtime path on every call — class-version check, a three-way inline-cache walk with its LRU reshuffle, frame build — 97 instructions, against 7 for `o.foo(...)` and 3–4x the time.

The name is in hand at bytecodegen, so such a site now carries a twin: a second call site describing `recv.foo(...)` — same receiver, the arguments past the name (already contiguous from `args + 1`), same result slot, and `bypass_visibility`, because a `send` reaches a private method and so must the twin. No bytecode executes it. `compile_method_call` takes it when the site's resolved target is the builtin `send` and the name resolves for the receiver class, and re-enters itself with the twin — so from there the target gets the ordinary treatment: specialization, ISeq-hint folding, the frameless expansions.

That the VM never looks at the twin is what keeps the semantics. Redefining `send` moves the class version; the repair re-asks this site's question, re-resolution no longer reaches the builtin, and the rewrite stops firing. The recursive call records its own inline-cache entry for the resolved name, so redefining *that* is caught the ordinary way. Two shapes decline: a set-guarded dispatch arm (the classes it admits agree on `send` and say nothing about what `:foo` resolves to), and a site whose PMC already holds more than one receiver class. A receiver-guard miss keeps the site's ordinary policy — under `Learn` it recompiles once, and the recompile reads the *send* site's PMC, by then holding the second class, so the test declines and the site settles on the lookup.

Only the plain shape gets a twin: literal Symbol name, no splat, no keywords, no `...`, no `&blk`. A literal block is forwarded (`enumerable.rb`'s `src.__send__(:__gather_each) { ... }` is the shape that matters there). `public_send` has no twin — its visibility check is the point of it.

5M calls, min of 5 interleaved runs:

| | before | after | the same call written out |
| --- | ---: | ---: | ---: |
| `o.send(:foo)` | 30.0 ms | **8.8 ms** | 9.0 ms |
| `o.send(:bar, 1, 2)` | 43.9 ms | **11.5 ms** | 11.8 ms |
| `o.send(m)`, name in a variable | 29.5 ms | 30.3 ms | — |

The site emits 97 → **7** instructions, the same as the direct call; the variable-name path is untouched and still emits its 95. Enumerable's `chunk` / `chunk_while` / `slice_when`, which reach their source through `__send__(:__gather_each)`, run at 0.89–0.94x.

## Verification

- `cargo test --release -p monoruby`: 2382 passed — the suite plus the two new test functions — with a failure set identical to the branch point's, before and after. (This container's CRuby is 3.3.6 where the tree wants 4.0.2, so the oracle-backed tests that depend on 4.0 behaviour fail here either way; they are the same 37 on master.)
- New tests: private / protected / accessor targets through `send`, a block, a polymorphic site, `method_missing`, `send(:send, ...)`, `BasicObject#__send__`, an arity error, a receiver class that flips after the site is compiled, and a redefinition of the target and of `send` itself. All pass on x86-64 and on aarch64 under qemu.
- Differential runs against CRuby outside the suite: `define_method` targets, Struct accessors, `Class.send(:new)`, refinements, backtraces through `send`, PIC formation over three receiver classes; and for part 1, `Hash#[]` over both regimes, default value / default proc, identity-keyed maps, frozen String keys, a subclass overriding `#default`, and `String#setbyte` over shared, frozen and out-of-range receivers.
- Re-verified after merging master (79 commits, including #1339's `super` change and the `Class#new` work in #1328 / #1334 / #1337, which touch the same dispatch code): the instruction counts above are unchanged (`send(:foo)` 7, `Hash#[]` 53 / 70, `setbyte` 35), the `object_send*` tests pass, and the differential scripts still match CRuby.

## Found along the way (filed separately, not in this PR)

- #1342 — `puts` raises `EPIPE` (stderr: SIGABRT) when stdout/stderr is a `/dev/null` opened `O_RDWR`, which is how `subprocess.DEVNULL` and several CI harnesses open it.
- #1343 — `public_send` skips `method_missing` for a private method.
- #1344, #1345, #1346 — three items the docs already record: no `TracePoint`, green threads stalling on kernel-blocking FFI, and `super` using the first-registered name when one proc is `define_method`'d under several names.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CSXFWf6MLpoTs9z98YXUTw)_


sisshiki1969:
CI on `ff7f02d`: `amd64` ✅, `darwin` ✅, `codecov/patch` ✅ — `codecov/project` ❌ at 92.95 % (−0.01 %). That drop is not from this diff, and here is the per-file evidence from Codecov's own compare API (`Δmiss` = new uncovered lines, head vs. base `3dd7e1a`):

| Δmiss | Δlines | file |
| ---: | ---: | --- |
| +14 | 0 | `hashbrown/src/raw/mod.rs` |
| +6 | 0 | `hashbrown/src/map.rs` |
| +3 | 0 | `monoruby/src/codegen/runtime.rs` |
| +2 | 0 | `monoruby/src/scheduler.rs` |
| +2 | 0 | `monoruby/src/builtins/io.rs` |
| +1 | 0 | `monoruby/src/value/rvalue/io.rs` |
| +1 | 0 | `monoruby/src/executor.rs` |
| 0 | +126 | `monoruby/src/builtins/kernel.rs` |
| 0 | +23 | `monoruby/src/bytecodegen/encode.rs` |
| 0 | +18 | `monoruby/src/bytecodegen/method_call.rs` |
| 0 | +5 | `monoruby/src/codegen/arch/x86_64/compile/builtin.rs` |
| 0 | +1 | `monoruby/src/bytecodegen.rs` |
| 0 | +1 | `monoruby/src/builtins/object.rs` |
| **−1** | +25 | `monoruby/src/codegen/jitgen/compile/method_call.rs` |
| **−3** | +11 | `monoruby/src/globals/store.rs` |

Every file the PR touches has `Δmiss ≤ 0` — the 210 added lines brought no new uncovered line, and two of those files ended up with fewer. The +25 net misses are 29 new misses in files with **zero line change** (hashbrown's raw table and map, the scheduler, the IO paths, `runtime.rs`), which is the coverage run landing differently on timing- and iteration-order-dependent code, minus the 4 my files removed. `codecov/patch` agreeing ("all modified and coverable lines are covered") is the same fact from the other side.

I have not re-run anything for it. The failing signal is an external status computed from the `amd64` job's coverage upload, not an Actions job of its own, so "re-running" it means re-running the whole ~15-minute `amd64` job to re-roll the same dice; the per-file numbers above already settle the question more firmly than another sample would. Happy to trigger that re-run if you would rather see it flip. For reference, #1339 and #1340 were both merged with `codecov/project` red and `codecov/patch` green, so this looks like the usual drift rather than a gate this PR broke.

Nothing else is open: no merge conflict (`mergeable_state: unstable` is exactly this one red check), and no review comments.

---

Issue #1185, opened 2026-08-26
Compile the ensure chain into the specialized non-local exit instead of falling back to the VM


sisshiki1969:
Optimization follow-up to #1180 (from its review discussion).

## Current state

Since #1180, a `break` / non-local `return` whose unwind path crosses an exception handler is lowered to the **generic** exit (`err_block_break` / `err_method_return`): the VM interprets the `ensure` bodies, the chain-deopt walk converts the suspended frames, and — for `break` — the defining frame resumes in the interpreter until the next `loop_start` re-enters the compiled loop via OSR. That is correct, but it pays a deopt on a path the compiler actually knows everything about: when the whole chain is inlined into one unit, the handlers it crosses are in-unit iseqs too.

The compile-time split is already in place — `in_protected_region` (the exiting instruction's own coverage) plus `check_exception_handler` (the suspended frames' coverage) decide the lowering, so the handler-free case keeps the specialized teardown and pays nothing. This issue is about the *other* side of that split.

## Proposed

Model the unwind edge instead of escaping to the VM: at the exit site, splice the `ensure` bodies the unwind passes through — innermost first, each compiled in its own frame's context — and then perform the existing specialized teardown (`lea rbp += Σ; leave; ret`). No `handle_error`, no chain conversion, no OSR re-entry; for `break`, the defining frame's compiled continuation keeps running.

This also fixes the modeling gap at the root rather than routing around it: the `ensure`'s writes to outer locals become visible to the abstract interpreter (the exact blind spot behind #1179's original mis-analysis), because the edge exists in the CFG.

## What the design has to handle

1. **Control-flow override semantics.** A `break` / `return` / `raise` *inside* an `ensure` replaces the in-flight non-local exit (CRuby semantics; the VM implements this with `defer_unwind` / `EnsureEnd` / `discard_deferred_unwind`). A spliced ensure containing another non-local exit either needs the same replacement semantics compiled in, or — simplest first cut — refuses splicing and keeps the generic lowering (`in_protected_region` of the ensure body itself, recursively).
2. **`$!` save/restore.** An exit leaving a rescue *clause* restores `$!` to the region-entry save (`restore_errinfo_on_exit`). Spliced exits crossing such a region need the equivalent emitted.
3. **Continuation-state merge.** For `break`, the defining frame's post-call state must become the merge of the normal-return edge and the new break edge (post-ensure). Today only the former exists; the merge is what makes the ensure's outer-local writes sound by construction.
[...]

PR #1367, opened 2026-09-16
Read a protected region's entry $! save straight off the Executor

sisshiki1969:
Follow-up to #1366 — the other half of the `ensure` surcharge on the normal path, and the larger half.

## The call

A protected region saves `$!` at its entry so a non-local exit leaving a `rescue` clause can put it back. That save went through the generic hooked-global runtime call, **once per invocation of every method carrying a `begin`..`ensure` or `begin`..`rescue`**:

```text
:00001 %2 = $(errinfo)
      mov  rdi,rbx / mov rsi,r12 / mov edx,0x6a      before
      movabs rax,<runtime::get_global_var> / call rax

      mov  rax,QWORD PTR [rbx+0x198]                 after
```

`$!` is a plain `Value` field of the `Executor` and `$(errinfo)`'s hook is `Some(vm.errinfo())`, so `AsmInst::LoadErrinfo` is the same read — with the call's FP save set and GP flush gone with it.

## Why it is equivalent, and why only this name

The load is sound only because `rbx` is the **current** `Executor`, which is what `$!` is per: CRuby keeps errinfo per execution context and so does monoruby, so a Fiber or a Thread reads its own. That is exactly what the runtime call did with its `vm` argument.

Only the internal name is specialized. `$(errinfo)` is what bytecodegen emits and is **not a name Ruby's parser can produce**, so no program can alias it, `trace_var` it, or otherwise put a hook in the way. User-written `$!` reads keep the generic path. (`trace_var` fires on assignment rather than on read in any case, and the store side is untouched here.)

The name is pre-interned as `IdentId::GVAR_ERRINFO_INTERNAL` so the recognition is an integer compare rather than a lock and a string hash.

## Measurements

| | master (with #1366) | this PR |
|---|---|---|
| hot method with a trivial `ensure` (3 runs) | 0.673 / 0.675 / 0.663 s | 0.486 / 0.482 / 0.475 s |
| same method, no `ensure` | 0.393 s | 0.393 s |

About 28%. Together with #1366's gate, the surcharge for wrapping a hot method's body in `begin`..`ensure` falls from **+63% to +23%** over the same method without one, and that method's compiled body shrinks from 209 to 186 bytes.

The find-first-with-cleanup shape from #1185 is unchanged (0.880 s → 0.877 s): its cost is the unwind, not the region entry. `app_fib`, `app_aobench`, `binarytrees`, `so_nbody`, `so_mandelbrot`, `quick_sort`, `tarai`, `bf`: unchanged in time, byte-identical output.

## Tests

`cargo test`: 140 test binaries, all pass. New `tests/errinfo_inline_load.rs` (5 tests) pins what the inline read has to preserve, with the region hot enough to be compiled: the entry save and its restore with a non-nil `$!` in flight, a nested `rescue` inside the region restoring the outer `$!`, a `rescue`-only region binding `$!` to its own exception, and — the reason the read is only sound off the *current* `Executor` — the same across a Fiber and across a Thread.

`doc/exception_handling.md` gains §6.2.

aarch64 mirrors the x86-64 load (`ldr x0, [x19, #EXECUTOR_ERRINFO]`) but could not be compile-checked here — it needs the macOS/arm64 CI leg.

---


sisshiki1969:
`codecov/project` is red at 93.02% (-0.01%). It is not this PR's: every line this PR adds is covered, and the eleven new misses are all in files whose line count did not change at all.

From the Codecov compare API for this PR (`a33238b` → `f1a5af2`), the only files with a coverage delta:

| file | misses | hits | lines |
|---|---|---|---|
| `hashbrown/src/raw/mod.rs` | +14 | −14 | **+0** |
| `monoruby/src/builtins/io.rs` | +2 | −2 | **+0** |
| `monoruby/src/builtins/kernel.rs` | +1 | −1 | **+0** |
| `hashbrown/src/map.rs` | −6 | +6 | **+0** |
| `codegen/arch/x86_64/compile/mod.rs` | +0 | +4 | +4 |
| `codegen/jitgen/asmir/compile_shared.rs` | +0 | +4 | +4 |
| `codegen/jitgen/compile/variables.rs` | +0 | +5 | +5 |
| `id_table.rs` | +0 | +2 | +2 |

`+14 +2 +1 −6 = +11`, which is exactly the reported miss delta. The four files this PR touches contribute `+15` lines and `+15` hits — zero new misses, which is also what the green `codecov/patch` ("All modified and coverable lines are covered by tests") says.

The four files that moved are ones this PR does not touch, with no line-count change, so this is run-to-run variance in which lines the suite happens to execute — the vendored `hashbrown` generics dominate it (14 of the 11 net), and `io.rs` / `kernel.rs` contribute three more. There is no fix to port; the threshold is Codecov's default of 0% drift on `project`, and the drift here is 0.01%.

Re-running `amd64` once to regenerate the coverage upload.

---

PR #1369, opened 2026-09-16
Keyword binding without the per-call clones, and the railsbench re-measurement after #1365

sisshiki1969:
Two things: the railsbench re-measurement on master after #1365 (§8 of `doc/railsbench_investigation_2026-09.md`, three commits), and the first fix it ranks, S's keyword half.

## Keyword binding without the per-call clones (`codegen/runtime/args.rs`)

The generic argument path (`set_callee_frame_arguments` → `handle_keyword`) cloned the whole `CallSiteInfo` — its splat and hash-splat Vecs, its keyword `IndexMap`, its polymorphic cache — on every keyword-passing call, to read one integer and one Vec out of it, and cloned `kw_args`, `hash_splat_pos` and the callee's keyword-name Vec again in the loops. Each clone was there because the loop bodies re-enter Ruby (`#to_hash`, `#hash` / `#eql?` on keys) and need `globals` mutably. On railsbench that was 76 calls a request at 1.2k instructions each (§8.4.3).

- The loops now index the call site and the callee and re-borrow per step, so nothing is cloned: the excess-keyword Hash for a keyword-less callee, `coerce_hash_splat_args`, `hash_splat_and_kw_rest`, and the invoker's keyword loop.
- `handle_keyword` returns at once for a callee with neither keyword parameters nor `**kwrest`: the only way it is reached for such a callee is a call passing no keywords, so its three helpers had nothing to do (the `method_missing(name, *args)` shape).
- Merging a `**hash` into `**kwrest` still works on a snapshot (an insert may call a key's `#hash`, which could mutate the source) but skips the keys bound to declared keyword parameters while iterating instead of removing them from the snapshot first; a Symbol key matches a parameter name by identity either way.

**Measured** (callgrind difference method, WARM=1,000, N=100/600):

| | Ir/req |
|---|---:|
| master `2f1bd40e` | 2,937,075 |
| this branch | **2,864,768 (−72,307, −2.5%)** |

The −72k: `CallSiteInfo::clone` −8.7k, `IndexMap::clone` −5.2k, malloc/free −24k, `handle_keyword` −4.9k, `handle_invoker_arguments` −4.1k. One `config.x.y` read through `OrderedOptions#method_missing` (`mmprobe.rb`): 9,061 → 8,623 instructions; the rest of that path is positional binding, not keywords. `benchmark/railsbench_check.rb` passes 15/15.

**Tests**: the existing keyword / splat / forwarding tests, plus `keyword_binding_sources` against CRuby with the JIT warm: literal pairs with one and two `**hash` splats, declared keywords plus `**kwrest`, `**kwrest` alone (a fresh Hash, not the splatted one), a keyword-less callee reached through a splat with trailing keywords, the missing-before-unknown error order, a String key as an unknown keyword, `**nil`, and `Class#new` forwarding into an `initialize` with keywords. Full workspace suite: 2,462 pass, the one failure being the pre-existing `float::tests::angle` (`1.123456789.ceil(15)` against the locally installed CRuby 4.0.6; green on CI's Ruby and on master's binary). `cargo check --target aarch64-unknown-linux-gnu` is clean.

A pre-existing limitation is documented rather than fixed: bytecodegen lays literal `k: v` pairs out ahead of every `**hash`, so `n(**h, b: 2)` yields `{b: 2, a: 1}` where CRuby keeps source order. Fixing it needs the parser to record the interleaving.

## Re-measurement on master after #1365 (§8 of the investigation)

Same method as §6/§7. Instructions are 0.97× CRuby+YJIT (2,937,075 vs 3,038,397 Ir/req), steady-state wall clock 1.19× (1.35 vs 1.13 ms/req; was 1.4×), objects unchanged at 1.71× (1,569 vs 919), RSS growth 0.1 KB/req (was 1.2), I1 misses +21% over CRuby (was +33%). The ranked remainder with evidence (§8.4) and candidates R–X with expected gains (§8.5): the two `gsub!(regex, Hash)` calls a request (134k), the `method_missing` config chain (17/req at 9,061 Ir vs CRuby's 5,322, measured in isolation), keyword marshalling (this PR), the allocation sources behind the 1.71× (runtime and rest Arrays 396/req, `Symbol#to_s` clones, `$~` MatchData), string interpolation (55/req at 1.45k), runtime method lookups without site caches, and polymorphic sites that never heal because boot-time classes fill the PMC's four ways. New probes (`gsubprobe.rb`, `mmprobe.rb`), the differential profile-statistics method and the `deopt`-feature exit-log grouping are in §8.6.

---


sisshiki1969:
`codecov/project` is red by −0.02% (25 lines), and I am leaving it: none of the lost lines are monoruby logic this PR touches. From codecov's per-file line reports for base `30a58ded` vs head:

- `hashbrown/src/raw/mod.rs`, 14 lines: the `rehash_in_place` branch of `reserve`, which was only reached through the per-key `remove` on the `**kwrest` snapshot that this PR replaces with a skip while iterating.
- `monoruby/src/value.rs`, 7 lines: `impl Display for Value` / `RV`, which was only reached through the `unknown keywords: ::b` message bug this PR fixes (the Symbol's display supplied the extra colon).
- `hashbrown/src/map.rs`, 6 lines: hashbrown's own `#[cfg(test)]` clone helper (`panic_in_clone`), unrelated.
- `rubymap/src/lib.rs`, 1 line: a closing brace of the `remove` path above.

Patch coverage is 100% as of `ab47d56b` (the String-key arm of the invoker's unknown-keyword message now has a test). `amd64` and `codecov/patch` are green; `darwin` is running.

---

PR #1370, opened 2026-09-16
Replay a frame's own ensure bodies inline for break and non-local return

sisshiki1969:
Step 2 of the `ensure` surcharge work (#1185), on top of #1366 / #1367 / #1368.

## The gap

A `break` out of a block and a non-local `return` were the last exits that left their **own** frame's `ensure` bodies to `handle_error`: the unwinder found the covering region, deferred the exit, ran the body interpreted, and re-delivered it from `EnsureEnd`. That is precisely the cost #1185's splice machinery (§6.3) was built to compile around.

A local `return` has never done this — `emit_ret` replays the bodies inline through `gen_all_pending_ensures` and exits clean.

## The change

Those two exits now replay their own frame's open regions inline as well: same set, same order, same `$!` protocol. With the bodies emitted ahead of the exit there is nothing left to compile around — the exit crosses no region, the JIT lowers it to the plain specialized teardown, and no deferral is created.

```ruby
arr.each do |x|
  begin
    break x if x == t
  ensure
    $n += 1
  end
end
```

The exit value is generated *before* the replay and popped *after* it, so the bodies take their temps above it and the exit instruction's recorded `sp` is unchanged. Raising it by the value's own slot changed JIT liveness even for exits that replay nothing, and cost about 4%.

## What keeps the bodies from running twice

The replay-span machinery #1368 just added, reused rather than duplicated. `emit_nonlocal_exit` extends the spans the replay just recorded over the **exit instruction itself**. `handle_error` is handed exactly that pc and asks the same "which regions are in force here?" question the raise path asks, so:

- `covering_ensure` → nothing to run,
- `errinfo_restore_slots` → nothing to restore,
- `nonlocal_exit_needs_vm_unwind` → `false`,
- `try_splice_exit` (which starts from `single_covering_ensure`) → declines on its own.

No second table, and no third place to keep in sync. The first draft of this change did carry a separate `bare_nonlocal_exits` table with three explicit gates; folding it into the spans removed all of it, along with a `BcIndex: Ord` impl it needed.

Two consequences of routing it through the spans, both in the diff:

- a span is now recorded even when the body generated no code — an `ensure nil end` still has a region the unwinder would otherwise run through the VM. The empty span is inert until an exit extends it. (This is what the trivial-body benchmark below measures; without it that shape saw no improvement at all.)
- `errinfo_restore_slots` takes the same replay-span cut the region lookups take. It must: a replayed body is preceded by its region's `$!` restore, so restoring again on the way out would undo whatever the body did to `$!`. It cannot go through `active_entries` — it is keyed on rescue *clause* spans — so it applies the filter directly via the shared `is_replayed_at`.

## Measurements

The `break`-inside-its-own-`begin`..`ensure` shape, 20 000 iterations over a 40-element array with 64 targets. A **trivial** ensure body, so what is timed is the machinery and not the body:

| | no `ensure` | with `ensure` | surcharge |
|---|---|---|---|
| master (`a1aa2ec1`) | 0.185 s | 0.389 s | 2.10× |
| this PR | 0.185 s | 0.207 s | **1.12×** |

With a body that does real work (`$n += 1`), the remaining gap is the body itself: 0.690 s → 0.605 s.

Standard benchmarks are unchanged (best of 2, seconds):

| | app_fib | quick_sort | binarytrees | so_nbody | app_aobench |
|---|---|---|---|---|---|
| master | 0.109 | 0.441 | 0.143 | 0.201 | 4.482 |
| this PR | 0.109 | 0.449 | 0.141 | 0.204 | 4.505 |

optcarrot: checksum 59662, 237 fps.

## Tests

New `tests/nonlocal_exit_inline_ensure.rs` — 9 tests counting body runs, so a double run fails them: both exits, nested regions innermost-first, a region with both `rescue` and `ensure`, an exit out of a `rescue` clause (the `$!` restore), a raising body that replaces the exit, a `break` written *inside* the replayed body, the exit value surviving the replay, and the no-region case.

`cargo test`: 144 test binaries, 4112 tests, all pass. The 9-shape raising-`ensure` battery from #1368 still matches CRuby 4.0.2 exactly.

`doc/exception_handling.md` gains §6.5, and §6.4 is corrected where this change extends it.

---


sisshiki1969:
## `codecov/project` is red (-0.06%), and the reason is worth a decision

`codecov/patch` is 100% — every line this PR adds is covered. The project delta is **-68 hits on 235 files**, and it is not noise: it is this change switching off machinery it no longer needs.

### Where the hits went

From the Codecov compare API, per-file hit deltas (files with no line change):

| Δ hits | file |
|---:|---|
| −23 | `codegen/arch/x86_64/compile/mod.rs` |
| −16 | `codegen/runtime.rs` |
| −10 | `codegen/jitgen/compile.rs` |
| −5 | `codegen/jitgen/asmir/compile_shared.rs` |
| −4 | `codegen/jitgen/context.rs` |
| −3 | `globals/error.rs` |
| −26 / −6 | `hashbrown/src/raw/mod.rs`, `hashbrown/src/map.rs` |
| −2 / +5 | `builtins/io.rs`, and four small positives elsewhere |

The vendored hashbrown pair (−32) is the usual run-to-run noise. The other **−61 is real and is exactly the splice / deferral / `EnsureEnd` lowering** that a `break` or non-local `return` inside its own frame's region used to go through.

### What actually stopped running

I instrumented `try_splice_exit` and ran the whole suite on both revisions:

| | master `a1aa2ec1` | this PR |
|---|---:|---:|
| `try_splice_exit` calls | 1570 | 1581 |
| **`SplicePlan::SameFrame` selected** | **304** | **0** |
| refused (`hosts == 0`) | 702 | 1006 |

So **#1187's stage-1 same-frame splice is now unreachable**, suite-wide. That follows from the design and is not an accident: `try_splice_exit` starts from `covering_ensure(bc_pos)` at the exit's own pc, and this PR's whole point is that the regions there are already settled. Stage 2 (`SplicePlan::Outer`, #1363) is untouched — it keys off the *host* frame's suspended call pc — and still fires 269 times.

It is also the right trade. The splice made an unavoidable deferral cheaper; replaying inline removes the deferral altogether, which is why the surcharge drops from 2.10× to 1.12× rather than partway.

### The decision

Two consequences, neither of which I want to settle unilaterally:

1. `SplicePlan::SameFrame`, its `emit_spliced_exit` arm, the `host_pos == current_pos` branch of `try_splice_exit`, and `AsmInst::DeferSplicedExit` become dead. (`record_spliced_ensure`, `spliced_ensures` and the `EnsureEnd` dispatch arms stay live — stage 2 uses them.)
2. `tests/spliced_exit_void_temp.rs` (the 4 tests from #1368) still pass but no longer reach the path they document; I confirmed the crash they pin can no longer be produced through those shapes.

Options:

- **(a) Merge as is.** The dead entry point stays, harmless, and the coverage dip stands. Those 4 tests stay as a guard that would fire again if the inline replay were ever disabled.
- **(b) Delete stage 1 in a follow-up PR.** A clean, narrow deletion — the delivery machinery is shared with stage 2 and stays. Coverage recovers because the unexecuted lines are gone rather than merely unreached. `spliced_exit_void_temp.rs` would be re-pointed or retired with a note.

I lean towards (b) as a separate PR, keeping this one to the behaviour change. I have not widened this PR to do it.

---

PR #1384, opened 2026-09-17
A &:sym block yielded nothing raises "no receiver given"

### LINKED ISSUE #1380: Yielding no arguments to a &:sym block raises "wrong number of arguments" instead of "no receiver given"
Found while working on #1376. On master `57ad3066`, with CRuby 4.0.6 as the reference.

```ruby
def y0(&b) = yield
y0(&:to_s)
# monoruby: ArgumentError: wrong number of arguments (given 0, expected 1+)
# CRuby:    ArgumentError: no receiver given

:to_s.to_proc.call   # both: ArgumentError: no receiver given
```

The `Symbol#to_proc` body (`SYMBOL_TO_PROC_BODY_FUNCID`) is declared with one required parameter, so a yield of nothing fails the binder's arity check before the body can raise its own message. The exception class matches; only the message differs.

sisshiki1969:
Closes #1380.

```ruby
def y0(&b) = yield
y0(&:to_s)   # was: wrong number of arguments (given 0, expected 1+)   CRuby: no receiver given
```

The `Symbol#to_proc` body was declared with one required parameter, so a `yield` with no arguments failed the binder's arity check before the body could raise CRuby's message. The body now binds zero arguments (`define_builtin_func_variadic(.., 0)`) and raises "no receiver given" itself, which `:to_s.to_proc.call` already did.

What the proc reports is unchanged: `FuncInfo::is_symbol_proc_body` names the body, directly or behind a `define_method(&:sym)` proc-method (a separate FuncId wrapping it), and `arity` / `build_parameters` keep `-2` and `[[:req], [:rest]]` for it, with `:req` even under `parameters(lambda: false)`, as CRuby.

## Tests

`symbol_to_proc_zero_arguments`, against CRuby: `yield` with zero, one, two and an Array argument into `&:sym`, `call` / `[]` / `yield` with nothing, arity, parameters with and without the `lambda:` override, `lambda?`, `curry`, `instance_exec(&:sym)`, a nil receiver, and `define_method(&:sym)` called with zero and one argument plus its arity and parameters. The symbol, proc, method and `define_method` suites pass (148).

---


sisshiki1969:
Superseded by #1385, which fixes the same issue (#1380) and also covers the keyword-only receiver (`:to_s.to_proc.call(k: 1)` → `"{k: 1}"`), which this branch still rejects with "no receiver given". Closing in its favor.

---

PR #1393, opened 2026-09-17
Dynamically loaded native extensions: the mr_* ABI, and sqlite3 / zlib / zstd / psych / nokogiri moved out of the core

sisshiki1969:
## Summary

monoruby had no way to load native code at run time, so every C-extension stand-in (sqlite3, zlib, zstd, psych, nokogiri) and its bundled C library was compiled into the one binary. This PR adds a small monoruby-specific extension ABI and moves all five stand-ins out of the core into dynamically loaded extensions (`ext/`). The design study behind it is `doc/native_extension_loading.md` (options compared, the chosen order, the actual ABI, lessons learned).

The work, in the order it was planned:

1. **`doc/native_extension_loading.md`** — the design study (CRuby C-API shim vs. a monoruby-specific ABI vs. others, with a staged recommendation).
2. **Feature gating** — each stand-in behind a default-on Cargo feature (scaffolding; removed again at the end once nothing was left to gate).
3. **The ABI and loader** — `monoruby_ext_sys` (`MrValue` / `MrContext` / `MrApi` function table, `include/monoruby_ext.h` for C), `monoruby/src/ext.rs` (the table's implementation, the method trampoline, `ExtNative` payloads, the `dlopen` loader, `require "x.so"` resolution through `MONORUBY_EXT_PATH` / the binary's directory / `<install root>/ext`), and `monoruby_ext` (the safe Rust wrapper: `Ctx`, `Value`, `method!`, `native!`). `tests/native_ext.rs` builds an extension written in C against the header and checks every API entry through the binary.
4. **sqlite3** → `ext/sqlite3` (`libsqlite3_native.so`). The blocking open/close still park the green thread (`NativeOp::Ext`); `create_function` / `create_aggregate` call back into Ruby from inside `sqlite3_step` as before.
5. **zlib and zstd** → `ext/zlib`, `ext/zstd`. rubygems needs `zlib`, so `bin/install` builds the extensions and puts them in the install root's `ext/`.
6. **psych** → `ext/psych` (libyaml events dispatched to `Psych::Handler` through `funcall`).
7. **nokogiri** → `ext/nokogiri` (12 files, ~7k lines over `libxml2-src`), the one with the most involved object lifetimes: node identity through `_private` + the document's node cache, and SAX / XPath / XSLT / IO callbacks reaching Ruby through the running call's `Ctx`, with a raised exception stashed (`error_take`) and re-raised after the library returns.

The core no longer depends on `libsqlite3-src`, `libz-sys`, `zstd-safe`, `libyaml-safer` or `libxml2-src`; `ObjTy::NATIVE` is used only by `ext.rs`. The release binary went from 17.6 MB to 12.3 MB.

### ABI in one paragraph

An extension exports `Init_<name>(MrContext*)`. Every call into it gets a fresh `MrContext` (the `MrApi` table plus the VM's opaque pointers); methods are `fn(ctx, self, argc, argv, block) -> MrValue` with `MR_UNDEF` meaning "an exception is pending". Values are monoruby's own bits; a held `MrValue` is rooted through the temp stack, `gc_pin`, or the payload's `mark`. Classes with `MR_CLASS_NATIVE` get a payload-less instance from the generic allocator and the extension fills it with `native_set` (so the ABI has no alloc hook). Blocking work goes through `call_blocking`, which parks the green thread. An extension never links `monoruby` and never unwinds across the boundary. Extension `static`s are process-wide while `Init_` runs once per interpreter, so per-interpreter state (the classes it defined) is keyed by `Ctx::interpreter_id`.

### Build / test plumbing

- `cargo build` at the workspace root builds the `.so`s next to the binary, where the loader finds them. `bin/install` builds them and copies them into `<install root>/ext/`; `bin/spec` uses it.
- Tests build an extension through `tests::ensure_extension` into `target/ext/` (the outer cargo holds the main target dir's lock) and either register the directory in-process or pass it as `MONORUBY_EXT_PATH` to the spawned binary. `bin/test` / `bin/test-aarch64` build the five extensions next to the benchmark binary.
- `mr_str_ptr` / `mr_str_encoding` raise CRuby's "no implicit conversion of X into String", which nokogiri's error tests depend on.

### Merged from master

master's `SQLite3::Backup` (#1391) landed in the builtin this PR had already moved out, so it was ported into `ext/sqlite3` with the two differences the ABI imposes: a fresh `Backup` has no payload until `initialize` fills it with `native_set` (there is no alloc hook), and the two-connection type check uses the ABI's `type_of`, keeping the extension's "(expected SQLite3::Backup)" message. bcrypt and markly are new core builtins and stay in the core.

### Coverage

`codecov.yml` ignores `ext/**` and `monoruby_ext/**`. `bin/test` builds the extensions with the instrumentation flags stripped and the interpreter loads them with `dlopen`, so the report step collects none of their counters — Codecov read `ext/` as 94 of 7409 lines covered although the tests exercise all of it through the binary. The interpreter's own half of the ABI, `monoruby/src/ext.rs` (857 lines, 82%), stays measured. Project coverage is 93.00% against master's 93.17%: the stand-ins took their well-covered lines out of the measured crate, which dilutes the average once, and a 1% threshold absorbs that. Instrumenting the extensions is the way to undo the exclusion, and is recorded in the doc.

### Not in this PR

Step 4 of the plan (a per-extension `bundled` / `system` feature to link the platform's library instead of the vendored one) and step 5 (a CRuby C-API shim, only if it turns out to be needed).

## Test plan

- [x] `cargo check -p monoruby --tests` clean
- [x] `tests/native_ext.rs` (C extension round trip), `tests/sqlite3.rs` (24, including master's new `sqlite3_backup`), `tests/zlib_deflate.rs` (6), `tests/zlib_gzip.rs`, zlib unit tests pass
- [x] Every test of the sqlite3, zlib, zstd, psych, yaml_scalars and native_ext files re-run one process per test, the way nextest runs them — this is what caught `sqlite3_unsupported_callbacks_refuse` building no extension, which a shared-process `cargo test` had masked
- [x] `tests/nokogiri.rs`: 20 of 23 pass against the host nokogiri gem in the dev container; the 3 remaining diffs are Ruby 3.3 vs 4.0 `inspect` formatting there (backtick quoting in `NoMethodError`, `=>` spacing in `Hash#inspect`), and all 23 pass under CI's Ruby 4.0.2
- [x] psych parser / emitter primitives verified through the binary (the psych integration tests need `date`, absent in the container)
- [x] CI green on Linux/x86-64 and macOS arm64 (the aarch64 path builds the extensions with `--target`)


sisshiki1969:
The `amd64` check failed on `de17c5a` with one test, `monoruby::markly markly_render_html`. It is not this PR's, and it is not a one-off: the rendered attribute order is non-deterministic per process.

**What differs.** Only one element of the compared array, the `FULL_INFO_STRING` code block:

```
expected  <pre><code class="language-rb" data-meta="x">code
</code></pre>
actual    <pre><code data-meta="x" class="language-rb">code
</code></pre>
```

**Root cause.** comrak 0.55 collects the code-block attributes in a `std::collections::HashMap` and writes them by iterating it (`src/html.rs`):

```rust
let mut pre_attributes: HashMap<&str, Cow<str>> = HashMap::new();
let mut code_attributes: HashMap<&str, Cow<str>> = HashMap::new();
...
code_attributes.insert("class", code_attr.into());
code_attributes.insert("data-meta", meta.into());
...
write_opening_tag(context, "code", code_attributes.into_iter())?;
```

`HashMap`'s `RandomState` is seeded per process, so whenever two attributes land on the same tag the emitted order flips between runs. That only happens with `full_info_string` (or `github_pre_lang`) plus a fence meta string, which is exactly what this one assertion exercises. Running the snippet ten times under the built binary gave `class` first seven times and `data-meta` first three times.

**Why it is not this PR's.** The test, `src/builtins/markly.rs` and the comrak dependency all arrived with #1391; this PR touches none of them, and its `Cargo.lock` pins comrak 0.55.0 exactly as master's does (the lock differs from master's only by this PR's own workspace crates). The same single test failed on master's own commit for #1391, [run 35240511984](https://github.com/sisshiki1969/monoruby/actions/runs/35240511984) on `f613828`; the green run on `22a4f78` is the same coin landing the other way.

**No fix exists to port**, so nothing was pushed here. A proposed patch, for the markly work rather than this PR: comrak already offers the hook that bypasses its own tag writing — `plugins.render.codefence_syntax_highlighter`. A small `SyntaxHighlighterAdapter` whose `write_pre_tag` / `write_code_tag` emit the attributes in cmark-gfm's fixed order (`class`, then `data-meta`, then `data-sourcepos`) makes the output deterministic without touching the rest of the renderer. Upstream, the two maps want an order-preserving container.

Re-running the failed job once, as the failure is expected to pass on a re-run about two times in three.

---

PR #1440, opened 2026-09-20
Derive `Encoding.default_external` from the locale, as CRuby does

### LINKED ISSUE #1433: Encoding.default_external is always UTF-8: it does not follow the locale, so `p "い"` differs from CRuby under LANG=C
Noticed while diffing `String#rindex` output against CRuby in a container with no `LANG` set. Not a regression and not new code — `src/tests.rs` already documents it and works around it by spawning the reference CRuby with `-E UTF-8` — but there is no issue tracking it, so here it is.

CRuby derives `Encoding.default_external` from the locale (`Encoding.locale_charmap`, i.e. `nl_langinfo(CODESET)` after `setlocale(LC_CTYPE, "")`). monoruby reports UTF-8 unconditionally. The visible consequence is `String#inspect`, which in CRuby escapes any character not encodable in `default_external`:

```
$ env -u LANG -u LC_ALL ruby -e 'p Encoding.default_external, "い"'
#<Encoding:US-ASCII>
"い"

$ env -u LANG -u LC_ALL monoruby -e 'p Encoding.default_external, "い"'
#<Encoding:UTF-8>
"い"
```

Measured on Ubuntu 24.04 (only `C.UTF-8` generated), monoruby `0d744ba1` + CRuby 4.0.6:

| environment | `locale_charmap` | `default_external` | `p "い"` |
|---|---|---|---|
| (no LANG) | both `"ANSI_X3.4-1968"` | CRuby `US-ASCII` / mono **`UTF-8`** | CRuby `"い"` / mono **`"い"`** |
| `LANG=C.UTF-8` | both `"UTF-8"` | both `UTF-8` | both `"い"` |
| `LC_ALL=C` | both `"ANSI_X3.4-1968"` | CRuby `US-ASCII` / mono **`UTF-8`** | CRuby `"い"` / mono **`"い"`** |
| `LANG=en_US.UTF-8` (locale **not generated**) | CRuby `"ANSI_X3.4-1968"` / mono **`"UTF-8"`** | CRuby `US-ASCII` / mono `UTF-8` | CRuby `"い"` / mono `"い"` |

So there are two halves:

1. **`default_external` is not derived from the charmap.** monoruby's `locale_charmap` is right in the first three rows; nothing maps it onto `default_external`.
2. **`locale_charmap` trusts the environment variable rather than the system.** In the last row the locale does not exist on the machine, so `setlocale` falls back to `C` and CRuby reports `ANSI_X3.4-1968`; monoruby answers `UTF-8` from the name alone.

A third, downstream one: with `LANG` unset CRuby treats `-e` source as US-ASCII and rejects a multibyte character outright (`invalid multibyte character`), while monoruby reads it as UTF-8.

## Why this is not a one-line change

Wiring `default_external` to the charmap would change `String#inspect` escaping, `IO`'s default external encoding, `File.read`, `ARGF`, and the source encoding of `-e` — for every run in a `C` / `POSIX` locale, which is exactly how CI and `cargo nextest` spawn things. It would make monoruby *more* CRuby-like, but it moves a lot of expected output at once (the test harness's `-E UTF-8` workaround exists for this reason and would want revisiting at the same time). Worth doing deliberately rather than as a drive-by.

sisshiki1969:
Closes #1433.

`Encoding.locale_charmap` read the environment variables; nothing then mapped it onto `default_external`, which answered UTF-8 whatever the locale said. CRuby derives both from `nl_langinfo(CODESET)`, so a `C` / `POSIX` locale — how CI, containers and `cargo nextest` run things — starts the process on US-ASCII.

```console
$ env -u LANG -u LC_ALL ruby -e 'p Encoding.default_external, "い"'
#<Encoding:US-ASCII>
"い"
```

## 1. The charmap comes from the system

`newlocale(LC_CTYPE_MASK, "")` + `nl_langinfo_l(CODESET)`, falling back to the `C` locale when the system has no such locale. That last part is the issue's fourth row: a `LANG=en_US.UTF-8` the machine never generated leaves `setlocale` on `C`, and CRuby reports `ANSI_X3.4-1968` where reading the name alone said `UTF-8`.

The locale goes into a throwaway object rather than being installed process-wide the way CRuby's `setlocale(LC_CTYPE, "")` does — the codeset is the same either way, and libxml2 and SQLite keep the `C` locale they have always had. The spelling stays the platform's (glibc `ANSI_X3.4-1968`, macOS `US-ASCII`); both name US-ASCII to `Encoding.find`.

## 2. `default_external` follows it

Seeded at startup, overridable by `-E` / `-K` / `Encoding.default_external=` exactly as before. `Encoding.find("filesystem")` and `Encoding.find("locale")` follow along as they already did — the reference CRuby answers US-ASCII for `filesystem` under no locale on **both** CI platforms, macOS included, so this PR leaves that alias exactly as it was.

## 3. `#inspect` escaping follows `default_external`

The machinery existed for `Array` / `Hash`; this makes it what CRuby does.

- `String#inspect` and `Symbol#inspect` escape non-ASCII to `\uXXXX` when the **result encoding** — `default_internal` if set, else `default_external`, with US-ASCII standing in for a non-ASCII-compatible one — cannot show it.
- A symbol whose name that hits is quoted rather than rendered bare: `:あ` under a UTF-8 locale, `:"あ"` under a `C` one.
- The rendering is tagged with that result encoding rather than always US-ASCII, so `"abc".inspect.encoding` is US-ASCII under a `C` locale and EUC-JP under `-E EUC-JP`.
- `Kernel#p` goes through `rb_inspect` (a new private `Kernel#__inspect`), so an `#inspect` of the object's own making — `Exception`'s included — is escaped where CRuby escapes it, and only there: `puts`, `to_s` and interpolation are untouched.

The decision is cached on `Store`, because the renderers (`Value::debug`, `inspect_symbol`) are handed a `Store` and nothing more.

## 4. A `-e` script is read in the locale's encoding

So a multibyte character in one is a SyntaxError under a `C` locale, as the issue's third, downstream point notes. The encoding reaches prism as an injected magic comment rather than being recorded after the parse, because it has to *lex* in it — otherwise `__ENCODING__` would say US-ASCII while the bytes came through as UTF-8. The injected line is subtracted back out, so `__LINE__`, `__FILE__`, backtraces and error carets stay put. A `# encoding:` comment in the script still wins, and so does `-K`; `-E` correctly does not move it.

## The test harness

`run_test` and friends pin this side to UTF-8 (`Globals::new_test`), matching the `-E UTF-8` the harness already spawns the reference CRuby with. Unpinned, both sides escape identically — but every non-ASCII entry in `ruby_oracle.tsv` would have to be re-recorded for a property those tests are not about. **The oracle is unchanged** by this PR.

## A detour worth recording

Three commits here (`9a38627`, `efab469`, `fbe3cb5`) add a macOS-specific `filesystem` encoding rule and then remove it again; they cancel out. I misread the first darwin failure's `assert_eq!` output — `left` is the *expected* side, the reference CRuby — and spent three runs pushing monoruby away from an answer that already matched. What ended it was `19ee5e1`, which makes the assertion name the first differing line and say which side is which instead of printing two escaped blobs. That part stays.

## Verification

- New `tests/encoding_locale.rs` (7 tests): spawns the binary under six locale environments and requires the same answers as the reference CRuby run the same way — the charmap, all four `Encoding.find` aliases, the `#inspect` shapes and their encodings, `Encoding.default_external=`, `-E`, `-K`, the `-e` source encoding and its line numbering, and that a script *file* stays UTF-8 whatever the locale. Differential rather than literal, since the codeset spelling is per-platform and a host may or may not have `C.UTF-8` / `en_US.UTF-8` generated.
- A 44-case differential script (containers, `Struct` / `Data` / `Set` / `Range`, recursive containers, custom `#inspect` public / private / non-String, `BasicObject`, binary strings, exceptions, `puts` vs `p`, `-E EUC-JP`, a UTF-16LE `default_internal`) is byte-identical to the reference CRuby under no locale, `LC_ALL=C`, `LANG=C.UTF-8` and `LANG=en_US.UTF-8`.
- `cargo test -p monoruby --lib`: 2570 passed, 0 failed (on the current head, with master merged in).
- Integration: `encoding_locale`, `command_line`, `spec_conformance`, `ruby_bench_outputs`, `main_script`, `literal`, `stringio_encoding`, `exception_api`, `backtrace`, `psych`, `yaml_scalars`, `prism_bridge`, `string_bytes`, `name_error_api`, `no_method_error_message`, `type_error_names`, `kernel_warn`, `warning_gating`, `require`, `nokogiri`, `sqlite3`, `hexapdf_output`, `msgpack`, `yajl`, `markly`, `date_parse` — all green. The remaining integration targets and ruby/spec are left to CI: the session's disk could not hold every debug test binary at once.


sisshiki1969:
### darwin: `encoding_locale::locale_charmap_and_default_external_match_cruby` — still red after three fixes

The `amd64` scope is green. The `darwin` job fails on exactly one line of the probe, identically on all three heads so far:

```
line 5: cruby "UTF-8" vs monoruby "US-ASCII"
```

That line is `puts Encoding.find("filesystem")` under **no locale at all**. Every other line of the probe — the charmap, `default_external`, `locale`, `external`, all the `#inspect` shapes and their encodings, the `-E EUC-JP` block — matches on macOS. So the locale work this PR is about is fine there; what fails is the one answer that is per-*platform* rather than per-locale.

**What CRuby does.** `rb_filesystem_encindex` is the locale encoding on Linux but a fixed UTF-8 on macOS (`#elif defined __APPLE__`), and from the first `Encoding.default_external=` on, every platform tracks that instead, because `enc_set_default_encoding` re-points the `filesystem` alias on each assignment. Before this PR monoruby matched macOS here by accident — `default_external` was always UTF-8 and `filesystem` was aliased to it. Deriving `default_external` from the locale is what exposes the gap, so leaving it unasserted would ship a real macOS regression rather than hide a test nit.

**What I have tried**, each verified on Linux and each still red on darwin:

1. `9a38627` — the macOS seed resolved `Encoding::UTF_8` through the constant table; that constant is not registered yet so early in `Globals::new`, so the lookup fell back silently. Switched to resolving by name.
2. `efab469` — dropped the startup seed entirely and applied the platform rule at the query (`filesystem_encoding_value`), by which point the interpreter is running Ruby and the constant table is complete. `$DEFAULT_FILESYSTEM` now records only an explicit assignment.

After (2) the remaining paths are all platform-independent, and I have exercised them on Linux with the macOS arm forced on: `Encoding.find("filesystem")` gives UTF-8 under `LC_ALL=C` and follows a later `default_external=` to EUC-JP, while the normal arm still matches CRuby under all four locales. `--lib` is 2562 passed throughout.

**What blocks me.** With the resolver lazy, US-ASCII on macOS means the `cfg!(target_os = "macos")` arm is not being taken — and that I cannot reproduce or observe from this Linux session. `19ee5e1` therefore makes the assertion self-diagnosing: it names the first differing line and prints whether the build compiled that arm. The next darwin run will say whether the arm is compiled in at all, which decides between "the rule never runs" and "the rule runs and something later overwrites it".

I have not re-run the job: three successive runs on three different commits failed the same way, so this is not a flake, and a re-run would spend the one I have on an answer I already know. I will keep the PR watched and act on the next darwin result. If it shows the arm *is* compiled in, the honest next step is a pair of hands on a Mac, and I will say so rather than keep pushing changes I cannot verify.

---

sisshiki1969:
### Correction: I had the darwin failure backwards

My previous comment stated the direction of that diff confidently and wrongly. The self-diagnosing assertion (`19ee5e1`) reports:

```
first difference: line 5: cruby "US-ASCII" vs monoruby "UTF-8"
built with target_os = macos: true
```

`left` in that `assert_eq!` is the *expected* side — the reference CRuby. So it was **CRuby answering US-ASCII** for `Encoding.find("filesystem")` on macOS and **monoruby answering UTF-8**, the opposite of what I read the first blob-vs-blob failure as. The macOS arm was compiled in and running the whole time; the premise it rested on was wrong.

The reference Ruby follows the locale for the filesystem encoding on macOS exactly as it does on Linux, so the `__APPLE__` arm of `rb_filesystem_encindex` is not what this build ends up on. My special case was taking the one value that already matched and pushing it away from CRuby.

`fbe3cb5` removes it, along with the `$DEFAULT_FILESYSTEM` gvar it needed. `filesystem` is an alias of `default_external` on every platform — what the code did before this PR, and what both CI platforms' reference CRuby confirms. **Net effect on this PR's diff: the filesystem encoding is untouched by it.** The three commits in between were a detour and cancel out.

Verified on Linux against the reference CRuby: `Encoding.find("filesystem")` is US-ASCII under no locale and follows a later `default_external=` to EUC-JP, matching exactly, and the probe's line 5 now agrees on both sides. `--lib` 2562 passed; the 44-case differential is byte-identical under all four locales.

The lesson worth keeping is the one in `19ee5e1`: the assertion now names the first differing line and which side is which, instead of printing two escaped blobs. That is what ended this, and it stays.

---

Issue #1455, opened 2026-09-20
The broken-string guard is bound to UTF-8, not to the receiver's encoding — 12 methods never reject EUC-JP / Shift_JIS / Emacs-Mule, and two reject UTF-8 that CRuby accepts


sisshiki1969:
Found while sweeping the EUC-JP / Shift_JIS character walk for #1444. Measured on master `00b2cc80` against CRuby 4.0.6; nothing here is touched by that work.

CRuby refuses to do character work on a receiver that is broken in its own encoding — `rb_enc_check` / `str_check_enc` raise `ArgumentError` before the method runs. monoruby's equivalent check only knows UTF-8, so it fires in the wrong places in *both* directions.

## 1. No rejection at all for the non-UTF-8 encodings

With `"A\x80B"` force-encoded to each (`\x80` is invalid in all three):

| method | CRuby | monoruby |
|---|---|---|
| `#count("A-Z")` | `ArgumentError: invalid byte sequence in <enc>` | `2` |
| `#delete("A")` | same | `"\x80B"` |
| `#squeeze` | same | `"A\x80B"` |
| `#squeeze("A")` | same | `"A\x80B"` |
| `#tr("A","z")` | same | `"z\x80B"` |
| `#tr_s("A","z")` | same | `"z\x80B"` |
| `#split("")` | same | `["A", "\x80", "B"]` |
| `#split(" ")` | same | `["A\x80B"]` |
| `=~ /A/` | same | `0` |
| `#upcase` | `ArgumentError: input string invalid` | `"A\x80B"` |
| `#downcase` | same | `"a\x80b"` |
| `#swapcase` | same | `"a\x80b"` |
| `#capitalize` | same | `"A\x80b"` |

Identical for `EUC-JP`, `Shift_JIS` and `Emacs-Mule` — all 13 rows, all three encodings.

`#length`, `#chars`, `#reverse`, `#index`, `#center`, `#inspect` and `#dump` are *not* in this group: CRuby answers for those, and so do we (correctly, after #1444).

## 2. Over-rejection for UTF-8

The mirror image — two methods raise on a broken UTF-8 receiver where CRuby just works:

```ruby
"A\xFFB".succ          # CRuby "A\xFFC",  monoruby ArgumentError: invalid byte sequence in UTF-8
"A\xFFB".sub("A","z")  # CRuby "z\xFFB",  monoruby ArgumentError: invalid byte sequence in UTF-8
```

[...]

PR #1482, opened 2026-09-20
String#succ: carry in the receiver's own encoding (#1457)

### LINKED ISSUE #1457: String#succ ignores the receiver's encoding: it increments as a Unicode scalar and writes UTF-8 bytes into an EUC-JP / Shift_JIS / Emacs-Mule string
Found while sweeping the EUC-JP / Shift_JIS character walk for #1444. Measured on master `00b2cc80` against CRuby 4.0.6.

CRuby's `rb_str_succ` carries in the receiver's own encoding: it increments the last character, and when the result is not a character in that encoding it carries into the one before. monoruby increments a Unicode scalar and emits the UTF-8 of it, whatever the receiver is labelled.

```ruby
[0x00, 0x7F].pack("C*").force_encoding("EUC-JP").succ.bytes
# CRuby    [1, 0]              ← 0x7F has no successor in EUC-JP, so it carries
# monoruby [0, 194, 128]       ← U+0080 in UTF-8, inside a string tagged EUC-JP

[0x00, 0x7F].pack("C*").force_encoding("Shift_JIS").succ.bytes
# CRuby    [0, 161]            ← the next Shift_JIS character after 0x7F is 0xA1
# monoruby [0, 194, 128]

[0x00, 0x7F].pack("C*").force_encoding("Emacs-Mule").succ.bytes
# CRuby    [1, 0]
# monoruby [0, 194, 128]

[0x00, 0x80].pack("C*").force_encoding("EUC-JP").succ.bytes
# CRuby    [1, 128]            ← 0x80 is not a character, so the carry skips it
# monoruby [0, 129]
```

The Shift_JIS row is the clearest statement of the rule monoruby is missing: `0x80..0xA0` lead nothing in Shift_JIS, so the successor of `0x7F` is `0xA1`, not `0x80`.

The result is not merely a different answer — it is a string whose bytes no longer belong to its own encoding. `[0, 194, 128]` tagged EUC-JP is `valid_encoding? == false`, produced from a receiver that was valid.

## And the reverse, on a broken UTF-8 receiver

The same method raises where CRuby answers:

```ruby
"A\xFFB".succ   # CRuby "A\xFFC", monoruby ArgumentError: invalid byte sequence in UTF-8
"A\xFF".succ    # CRuby "B\xFF", monoruby ArgumentError
```

That half belongs with #1455 (the broken-string guard being bound to UTF-8) rather than here, but it is the same method, so fixing one without the other leaves `#succ` half-right.

## Shape of the fix

`rb_str_succ`'s rule is: find the rightmost alphanumeric character and increment it with carry within its class (`z`→`aa`, `9`→`10`); with no alphanumeric, increment the rightmost *character* and carry left when it overflows. "Overflows" is where the encoding enters — after #1444 each self-walked encoding has a `precise_mbclen`, so "is this byte sequence a character here" is answerable, and the carry is "increment until `precise_mbclen` says `Char`, or run out and carry".

Broken/edge-case behaviour; no ruby/spec example covers the non-UTF-8 side.

sisshiki1969:
Fixes #1457. Also the `#succ` / `#succ!` part of #1455 §2.

`#succ` incremented a Unicode scalar and wrote its UTF-8, whatever the receiver was labelled — so an EUC-JP string came back holding bytes that are not characters in EUC-JP, `valid_encoding?` false, from a receiver that was valid.

| receiver | before | after / CRuby 4.0.6 |
|---|---|---|
| `[0x00, 0x7F]` EUC-JP | `[0, 194, 128]` | `[1, 0]` — `0x7F` has no successor, so it carries |
| `[0x00, 0x7F]` Shift_JIS | `[0, 194, 128]` | `[0, 161]` — `0x80..0xA0` lead nothing there |
| `[0x00, 0x80]` EUC-JP | `[0, 129]` | `[1, 128]` — `0x80` is no character, so the carry skips it |
| `"A\xFF"` (broken UTF-8) | `ArgumentError` | `"B\xFF"` |
| `[0x7F]` US-ASCII | `[194, 128]` | `[1, 0]` |
| `[0x41, 0xFF]` BINARY | `[65, 196, 128]` | `[66, 255]` |
| `[0xFF]` ISO-8859-1 | `[1, 0]` | `[248, 248]` — `ÿ` wraps to the bottom of the Latin-1 lowercase run |

## What changed

CRuby's `str_succ`, ported over the receiver's own encoding: step the rightmost alphanumeric character within its class, carrying left through the alphanumerics; with no alphanumeric anywhere, step the rightmost *character*, carrying the same way. Both halves are where the encoding enters — "character" is `precise_mbclen`'s answer, so a byte that starts none is passed over rather than incremented, and a step that runs off the top wraps to the smallest character of the same byte width (`"߿".succ` is a carry in front of `"\u0080"`, not in front of two NUL bytes).

`enc_succ_char` / `enc_pred_char` / `enc_succ_alnum_char` / `enc_str_succ` are the four pieces, with `min_char_bytes` for the wrap. The alphanumeric classes are Onigmo's, which differ per encoding and which `#succ` makes visible:

- **UTF-8** uses the Unicode properties (`"Zあ".succ` steps the `あ`), including the spacing combining marks but not the non-spacing ones (`"wः".succ` steps the mark, `"w͙".succ` steps the `w`);
- **EUC-JP / Shift_JIS** count no multi-byte character as one, so `"Z\xA4\xA2".succ` wraps the `Z` and leaves `あ` alone;
- **Emacs-Mule** counts all of them;
- the **single-byte** encodings split: the ISO-8859 family, the KOI8 pair and the Windows code pages have letters in their high half, while the DOS code pages and the Mac encodings do not (`"\x9A"` in IBM437 just steps). TIS-620 marks every byte it defines alnum, `฿` and `๛` included.

Two things follow: a **broken** receiver answers instead of raising — the `#succ` / `#succ!` half of #1455 §2 — and BINARY / US-ASCII receivers stop stepping bytes their encoding has no character for. The byte-only `byte_succ_no_alnum` special case is subsumed and gone.

## Verification

- Differential run against CRuby 4.0.6 over **8300+ generated cases**: 300–400 random byte strings per encoding across 13 encodings (valid, broken and empty receivers, 0–4 bytes), the Unicode block edges (U+07FF, U+FFFF, U+10FFFF, the surrogate neighbours, full-width letters and digits), the general-category boundaries (Mn / Mc / Me / Nl / No / Lm / Lo), and the whole EUC-JP / Shift_JIS / Emacs-Mule / single-byte matrix from the issue — byte-for-byte identical, errors included. Two of those matrices are checked in as `succ_walks_the_receivers_own_encoding` and `succ_on_broken_and_byte_receivers`.
- `cargo nextest run --features stress-spill-pool` — 4180 passed, 0 failed.

`String#upto` and `Range#each` ride on `#succ`, so #1481's empty-endpoint fix and this one meet there; this branch is cut from master and the two do not conflict.

https://claude.ai/code/session_011eHp6aUJL4RoGPByjrszPv

---
_Generated by [Claude Code](https://claude.ai/code/session_011eHp6aUJL4RoGPByjrszPv)_


sisshiki1969:
`codecov/patch` was failing at 88.30% (40 lines). Pushed `8cee36a` with three matrices, each measured against CRuby 4.0.6 before being recorded, which takes the patch to **94.15%** locally (`cargo llvm-cov nextest -p monoruby`, cross-referenced against this branch's own diff), and merged master in (`fc083ac`) so the report is no longer six commits behind.

What the new tests reach that the old ones did not:

- **the walk down to the bottom of a class** (`enc_pred_char`). Below NKO DIGIT ZERO the two-byte UTF-8 tail has to borrow from the lead byte, which is the only way into that branch; and `"ª".succ` is the case where the class holds one character only, so `#succ` steps the character instead.
- **the top and bottom of Emacs-Mule's two-byte width** — stepping off the top wraps within the width rather than carrying, which is the `Neighbor::Wrapped` arm that returns `Found`.
- **the single-byte tables**, which were only being asked about ISO-8859-1, IBM437, KOI8-R and TIS-620: the ISO-8859 family's C1 controls, an undefined byte, a combining mark (Windows-1258 `0xCC`), and the two Windows pages whose script Onigmo does not mark alpha.

The 18 lines still uncovered are the deep arms of `enc_pred_char` and `min_char_bytes` — the ones a single-byte pred never reaches, since every byte of a single-byte encoding is a character and the first decrement always succeeds. They are there because the port follows CRuby's, and I would rather keep them than trim the port to what `#succ` happens to exercise. Say the word if you would prefer them gone.

**One thing the byte-by-byte measurement turned up that this PR does not fix**: `succ_alnum_class` derives the alphanumeric class from the Unicode category, where Onigmo carries a per-encoding table, and the two disagree on **29 bytes** out of 3456 across the single-byte encodings — Onigmo marks a contiguous run of its own script and nothing else, so a modifier letter (Windows-1252 `0x88` = U+02C6) or a letter outside the main run (Windows-1256 `0x8A` = U+0679) is not ALPHA for it. ISO-8859-11 is a third case: it has no codec here, so it reads as Latin-1 rather than Thai. Filed as #1488 with the full table, since closing it means either carrying Onigmo's tables in tree or narrowing the rule — too wide for this PR. I also confirmed the `Windows-1255` / `-1256` exclusion in `high_half_has_letters` is the *better* of the two approximations: with them included those encodings disagree on 4 and 11 bytes respectively, with them excluded on 2 and 3.

---

Issue #1476, opened 2026-09-20
Name and inspect strings are tagged UTF-8 where CRuby tags them US-ASCII


sisshiki1969:
Found while implementing #1447 (a `String#strip` differential test compared `…encoding.name` results and tripped on the tag rather than on the text). Measured on master `c1d101da` against CRuby 4.0.6.

CRuby builds the strings it *names* things with out of US-ASCII when their content is ASCII-only: `rb_usascii_str_new` for encoding and class names, and `rb_str_inspect`'s result takes the ASCII-compatible result encoding. monoruby tags all of them UTF-8. The text is identical — only `#encoding` differs:

```ruby
p Encoding::UTF_8.name.encoding      # CRuby #<Encoding:US-ASCII>, monoruby #<Encoding:UTF-8>
p "x".encoding.name.encoding         # same
p String.to_s.encoding               # same
p String.name.encoding               # same
p :sym.to_s.encoding                 # same
p 1.to_s.encoding                    # same
p nil.inspect.encoding               # same
p "a".inspect.encoding               # same
p "あ".inspect.encoding              # CRuby #<Encoding:US-ASCII> (the escapes are ASCII), monoruby UTF-8
```

All nine rows are `US-ASCII` in CRuby and `UTF-8` here.

It is invisible until the string is combined with another one, at which point the tag decides the result: a US-ASCII name yields to whatever it is appended to, while a UTF-8 one wins over a US-ASCII receiver and, since #1453, does so in `File.join` too. It also shows up directly in any test or code that asserts on `.encoding`.

The rule to implement is CRuby's: build these with US-ASCII when every byte is ASCII, and with the natural encoding otherwise (`"あ".to_sym.to_s` is UTF-8 in both today). The sites are the name accessors (`Encoding#name` / `#to_s` / `#inspect`, `Module#name` / `#to_s` / `#inspect`, `Symbol#to_s` / `#name`), the numeric `#to_s` family, and `String#inspect` / `#dump` — the last two already produce ASCII-only text by construction for a non-ASCII receiver, so their result is always US-ASCII in CRuby.

PR #1490, opened 2026-09-21
name the function CRuby's SystemCallError fails in (#1454)

### LINKED ISSUE #1454: Errno messages from File.delete / File.chmod / File.symlink carry the wrong syscall tag
CRuby tags a `SystemCallError` message with the C function that failed (`… @ <tag> - <path>`). Three `File` class methods disagree with it.

## Reproduction

```ruby
t = "/tmp/nonexistent_x"
[
 -> { File.delete(t) },
 -> { File.chmod(0600, t) },
 -> { File.symlink("x", "#{t}/y") },
].each { |b| begin; b.call; rescue => e; p [e.class, e.message]; end }
```

| call | CRuby 4.0.6 | monoruby |
|---|---|---|
| `File.delete` / `unlink` | `"No such file or directory @ apply2files - /tmp/nonexistent_x"` | `"No such file or directory - /tmp/nonexistent_x"` (no tag) |
| `File.chmod` | `"No such file or directory @ apply2files - /tmp/nonexistent_x"` | `"… @ rb_file_chmod - …"` |
| `File.symlink` | `"No such file or directory @ rb_file_s_symlink - (x, /tmp/nonexistent_x/y)"` | `"… @ rb_file_s_symlink - /tmp/nonexistent_x/y"` (only the second path) |

`File.delete`, `File.chmod`, `File.lchmod`, `File.chown`, `File.lchown` and `File.utime` all run through CRuby's `apply2files`, so they all report that tag. `File.symlink` uses `sys_fail2`, which renders **both** paths as `(old, new)`; `File.link` and `File.rename` take the same shape and are worth checking in the same pass.

The exception class and errno are already correct in every case — only the message text differs. `MonorubyErr::errno_with_msg` (no tag) is what `File.delete` uses today; switching it to `errno_with_path(…, "apply2files", …)` covers the first two rows, and `sys_fail2` needs a small helper of its own for the pair rendering.

Found while fixing the raw-path handling in these same functions; the raw bytes of a non-UTF-8 path now reach the message correctly, so only the tag is left.

sisshiki1969:
Fixes #1454.

The tag in `… @ <tag> - <path>` is the *implementation's* name, not the Ruby method's, so a whole family of methods reports one name. Measured across 38 failing calls in `File` and `Dir` against CRuby 4.0.6; every one of them now matches.

## apply2files

`File.delete` / `.unlink` / `.chmod` / `.lchmod` / `.chown` / `.lchown` / `.utime` all go through CRuby's `apply2files`, and all say so whatever they were called as. monoruby had three different answers and one missing one:

| call | before | after / CRuby |
|---|---|---|
| `File.delete` | `No such file or directory - <path>` | `… @ apply2files - <path>` |
| `File.chmod` | `… @ rb_file_chmod - …` | `… @ apply2files - …` |
| `File.chown` / `.lchown` | `… @ rb_file_s_chown - …` | `… @ apply2files - …` |
| `File.utime` | `… @ rb_file_s_utime - …` | `… @ apply2files - …` |

## Two paths, and the EEXIST exception

`File.symlink` / `.link` / `.rename` go through `syserr_fail2_in`, which renders **both** paths. monoruby named one — and a different one in each of the three (`new`, `new`, `from`):

```ruby
File.link(miss, "#{d}/l")
# before  No such file or directory @ rb_file_s_link - <d>/l
# after   No such file or directory @ rb_file_s_link - (<d>/no, <d>/l)
```

`EEXIST` is that helper's own exception: CRuby names the second path alone, under the tag `syserr_fail2_in` *itself*, since that branch takes the function name from the helper rather than from its caller. Odd, but reproducible:

```ruby
File.link(ok, ok2)   # File exists @ syserr_fail2_in - <d>/ok2
```

The new `errno_with_paths` covers both shapes and keeps a non-UTF-8 path's raw bytes the way `errno_with_path` already does.

## One-word corrections

`File.truncate` reported the open it does on the way (`rb_sysopen`) rather than itself; `File.readlink` used the Ruby method's name where CRuby uses `rb_readlink`; `File.lstat` borrowed `rb_file_s_stat`; `Dir.chdir` carried no tag at all (`chdir_path`).

## Two things the same measurement turned up

The issue's own list names both, so they are fixed here rather than filed:

- **`File.chmod` with no arguments aborted the process.** It indexed past the end of its argument list, so `File.chmod` panicked (`index out of bounds`) through the `extern "C"` boundary instead of raising `ArgumentError: wrong number of arguments (given 0, expected 1+)`.
- **`File.lchmod` did not exist** (`NoMethodError`), so it had no tag to fix. glibc's `lchmod` is `fchmodat` with `AT_SYMLINK_NOFOLLOW`, which is what CRuby ends up calling too, so the errno matches on either platform: the Linux kernel refuses a symbolic link with `ENOTSUP`, Darwin honours the flag.

## Verification

- Four probes over `File` and `Dir` — 38 failing calls covering every tag above, both `EEXIST` shapes, the arity guards and `File.lchmod`'s success path — byte-identical to CRuby 4.0.6, exception class and message both.
- Three regression tests, on `run_test_once_live` rather than the snapshot oracle: a filesystem errno is not the same everywhere (unlinking a directory is `EISDIR` on Linux and `EPERM` on Darwin), so these are exactly the environment-dependent case `CLAUDE.md` reserves the live helpers for.
- `cargo nextest run --features stress-spill-pool` — 4207/4207.

https://claude.ai/code/session_011eHp6aUJL4RoGPByjrszPv

---
_Generated by [Claude Code](https://claude.ai/code/session_011eHp6aUJL4RoGPByjrszPv)_


sisshiki1969:
`codecov/patch` was failing at 92.89%. Both uncovered places were a case this PR had not measured, and one of them was hiding a difference:

**`Dir.chdir` with a block fails one function further in.** `Dir.chdir(path)` reports `chdir_path`, but `Dir.chdir(path) { }` reports `dir_chdir0` — and so does the restore that runs after the block, when the directory it came from is gone:

```ruby
Dir.chdir(missing)        # No such file or directory @ chdir_path - …
Dir.chdir(missing) { }    # No such file or directory @ dir_chdir0 - …
Dir.chdir(gone) { Dir.rmdir(Dir.pwd) }   # … @ dir_chdir0 - …  (the restore)
```

Fixed in `bc49d22`, with all three in the test.

The other place — a two-path call whose paths are not UTF-8 — was already right and now has a test: the raw bytes survive on either side of the `(old, new)` pair and the message comes back `ASCII-8BIT`. That test reports the message's encoding only when it carries such a byte, since what an all-ASCII message is tagged with is #1476 (CRuby says UTF-8, monoruby US-ASCII) and not this PR's business.

Patch coverage is 168/168 locally (`cargo llvm-cov nextest -p monoruby`, cross-referenced against this branch's diff); the suite is 4214/4214. master is merged in, so the report is no longer two commits behind.

---

